PR #84, opened 2021-01-11
Update to full APACHE-2.0 license text

thorvaldj:

### DIFF
--- a/LICENCE
+++ b/LICENCE
@@ -1,13 +1,202 @@
-Copyright 2020 Equinor ASA
 
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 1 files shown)


LICENCE
```diff
@@ -1,13 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
marmid74, line 190: Legge til Equinor ASA

thorvaldj: No, this should not be neccessary. This appendix is the optional boilerplate text to be used for individual files. Instead replace these fields when used.

PR #124, opened 2021-02-03
[PH] Fix #100 Add dataVersion in GetCap request

### LINKED ISSUE #100: 🐛'dataVersion' is a required optionsIn parameter to WMLS_GetCap()
**:bug: Describe the bug**
When WE authenticate the user against a WITSML server, it do send a WMLS_GetCap () request to the server with the parameter 'optionsIn' set to 'returnElements=header-only'

As per WITSML API specifications, 'dataVersion' is a required optionsIn parameter to be passed as part of the WMLS_GetCap() . Not doing so would result in the server returning a -424- error status code [Cf. Screenshot 1], which is then interpreted by WE as a 500 error [Cf. Screenshot 2]

In order to address the defect, and assuming that at this stage only WITSML 1.4.1.1 is being supported, the file backend OptionsIn.cs must be enhanced in scenario where the request is made to serve WMLS_GetCap()

After doing the change locally (in Fork) the authentication succeed. However as 'dataVersion' isn't a parameter required for all WMLS function it might be necessary to decide how handle the settings.

**To Reproduce**
1. After installing WITSMl Explorer and configuring it with Schlumberger WITSML developer test server 
2. Click on Server list
3. Authenticate 
    Verify first that the authentication is valid with another WITSML application \ toolbox 
4. See error screen shot below

**Expected behavior**
1. In scenario where input parameters to teh WMLS_** function being called is invalid, display proper error message as per WITSML specification error table. Only treat error to be (500) as ultimate fallback (InvokeAsync --> ExceptionMiddleware.cs default exception catch is to treat error as 500)
2. Fix VerifyCredentials and TestConnectionAsync to pass proper optionsIn parameter to the server

**Screenshots**

1. 424 Error code status returned when OptionsIn missing 'dataVersion' required parameter
image

2. WE Error message after failure of authentication. 500 error code isn't matching teh server code being returned.
image

**Desktop (please complete the following information):**
  - Browser chrome
 - Version : Version 87.0.4280.141 (Official Build) (64-bit)

**Smartphone (please complete the following information):**
 - N/A

**Additional context**

Pascal62223:
Fixes # (100)

## Description
As per WITSML API specifications, 'dataVersion' is a required optionsIn parameter to be passed as part of the WMLS_GetCap() . Not doing so would result in the server returning a -424- error status code.

In order to be compliant, the WMLS_GetCapRequest as been modified so that a dataVersion attribute is always specified as part of the OptionIns parameter.

By default, the value 'dataVersion=1.4.1.1' is being used, unless the optionsIn parameter used to initialize the WMLS_GetCapRequest already contains a 'dataVersion' attribute value.

In order to do that, the previously public OptionsIn property of teh WMLS_GetCapRequest class has been made private and a property setter/getter be implemented (the setter taking care of providing the default value = 1.4.1.1)

...

## Type of change

- [x] Bugfix

*

# Checklist:

- [ x I have used descriptive naming on components, functions and variables to avoid additional explanatory comments in the code
- [x] I have made corresponding changes to the documentation

## Further comments
N/A

### DIFF
--- a/Src/Witsml/ServiceReference/WitsmlService.cs
+++ b/Src/Witsml/ServiceReference/WitsmlService.cs
@@ -204,17 +204,34 @@ namespace Witsml.ServiceReference
     public partial class WMLS_GetCapRequest
     {
         
-        [System.ServiceModel.MessageBodyMemberAttribute(Namespace="", Order=0)]
-        public string OptionsIn;
+        [System.ServiceModel.MessageBodyMemberAttribute(Namespace="", Order=0, Name="OptionsIn")]
+        private string optionsIn;
         
         public WMLS_GetCapRequest()
         {
         }
         
-        public WMLS_GetCapRequest(string OptionsIn)
+        public WMLS_GetCapRequest(string OptionsIn )
         {
-            this.OptionsIn = OptionsIn;
+            this.OptionsIn = OptionsIn ;
         }
+
+        /// <summary>
+        /// Quoting the specification : "'dataVersion' is a required optionsIn parameter to be passed as part of the WMLS_GetCap(). 
+        ///                             Not doing so would result in the server returning a -424- error status code"
+        /// The OptionsIn setter method enforce such rule by adding automatically the "dataVersion" parameter if not provided.
+        /// It is to be noted that the data version is fixed to 1.4.1.1 by default and might need to be passed as argument in future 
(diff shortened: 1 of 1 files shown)


steinsiv (approved):
Thanks! 
The version lock-in might be something we need to attend to,  I agree. 1.4.1.1 should be ok for now. Testet against two servers - Merging PR

PR #187, opened 2021-02-17
Fix DeleteCurveValuesWorkerTest to work with universal time

HakonHesselberg:
Fixes # (issue)

## Description
This test failed on my computer, maybe because I'm running Windows instead of Linux. Changing the test to use universal time works on both operating systems.
...

## Type of change

- [x] Bugfix

* Testing

# Checklist:

- [ ] PR is related to an issue
- [ ] New and existing unit tests pass locally with my changes
- [ ] I have used descriptive naming on components, functions and variables to avoid additional explanatory comments in the code

### DIFF
--- a/Tests/WitsmlExplorer.Api.Tests/Workers/DeleteCurveValuesWorkerTest.cs
+++ b/Tests/WitsmlExplorer.Api.Tests/Workers/DeleteCurveValuesWorkerTest.cs
@@ -128,13 +128,13 @@ namespace WitsmlExplorer.Api.Tests.Workers
             {
                 new IndexRange
                 {
-                    StartIndex = new DateTime(2000, 1, 2).ToString(CultureInfo.InvariantCulture),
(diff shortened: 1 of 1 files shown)


steinsiv (commented):
Good catch! 👍 

I think we probably should use ISO UTC format on all DateTime, (_ref. v1.4.1v3 WITSML APPENDIX B: object selection capability_)

I found an extension method at `Witsml.DateTimeExtensions.ToISODateTimeString()` . Should we expand on that instead?. 

ISO 8601 which I think the name refers to,  also allows formats with local time and UTC offsets, so maybe we need to clarify or implement those as well in the extension? 
Let me know your thoughts on this.

HakonHesselberg:
I think for this pull request the most important part is to just get the tests up and running. I will use the extension instead as it does the same as my code so we only need to change the extension if we want to improve upon that in the future.

Expanding on the extension to allow for local time and UTC offsets is a good idea, but should maybe wait until it is going to be used.

steinsiv (approved):
Thanks!

PR #263, opened 2021-03-15
Visualise Active Well and Wellbore

### LINKED ISSUE #232: 👉 Visualize Active WELL / WELLBORE
This task is related to User Story #177 

It is important for users to quickly and easily get an overview over which WELLS, WELLBORES and LOGS are active.

A visual indicator similar to what is seen in the picture: 
🟢 A green circle indicates the object or an object lower in the hierarchy is active. 
⚪ An empty circle indicates the object and no objects lower in the hierarchy are active.

Using a similar approach to what exists in current solutions makes it easier for users of WITSML Explorer to recognise the functionality.

image

Currently no alternatives, but open to suggestions

**Additional context**
The relevant WITSML data comes from the **isActive** tag.
```XML
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
<wellbores version="1.4.1.1" xmlns="http://www.witsml.org/schemas/1series">
	<wellbore uidWell=" uid=">
		<nameWell />
		<name />
		<isActive>true</isActive>
	</wellbore>
</wellbores>
```

HakonHesselberg:
Fixes #232 

## Description

This pull requests adds the isActive parameter to the list of parameters fetched when requesting data on a wellbore. This attribute is then passed to the front-end and displayed with a 'trending-upwards' arrow as seen in the picture. A well is also indicated as active if one or more wellbores are active.

image

...

## Type of change

- [ ] Bugfix
- [x] New feature (non-breaking change which adds functionality)

* Sidebar - added icon to active wells and wellbores
* Wellbore service - added requesting isActive parameter

# Checklist:

## Further comments

I have an issue with fetching the `isActive` parameter from Sitecom test server. I'm not sure if this is because there is nothing to get, but I would be happy to update the PR to correctly get the `isActive` parameter from any server which uses it.

If I add `[XmlElement("isActive")] public string IsActive { get; set; }` to `WitsmlWellbore.cs` 
as well as `IsActive = ""`, to the `QueryAll` in `WellboreQueries.cs` then I get true/false values as strings from Aker BP Test server and null values from Sitecom test server.
However if I make the `IsActive` parameter as a boolean then I get a true/false value as a bool from Aker BP Test server, and no wellbore results at all from Sitecom test server.

### DIFF
--- a/Src/WitsmlExplorer.Frontend/components/Icons/IsActiveIcon.tsx
+++ b/Src/WitsmlExplorer.Frontend/components/Icons/IsActiveIcon.tsx
@@ -1,0 +1,9 @@
+﻿import * as React from "react";
+
+export function IsActiveIcon(): React.ReactElement {
+  return (
+    <>
+      <img src="/assets/icons/trendingUp.svg" alt="Is Active" />
+    </>
+  );
+}
--- a/Src/WitsmlExplorer.Frontend/components/Sidebar/WellboreItem.tsx
+++ b/Src/WitsmlExplorer.Frontend/components/Sidebar/WellboreItem.tsx
@@ -118,6 +118,7 @@ const WellboreItem = (props: WellboreItemProps): React.ReactElement => {
       labelText={wellbore.name}
       onLabelClick={onLabelClick}
       onIconClick={onIconClick}
+      isActive={wellbore.isActive === "true"}
     >
       <TreeItem
         nodeId={logGroupId}
--- a/Src/WitsmlExplorer.Frontend/components/Sidebar/WellboreItem.tsx
+++ b/Src/WitsmlExplorer.Frontend/components/Sidebar/WellboreItem.tsx
@@ -118,7 +118,7 @@ const WellboreItem = (props: WellboreItemProps): React.ReactElement => {
       labelText={wellbore.name}
       onLabelClick={onLabelClick}
       onIconClick={onIconClick}
-      isActive={wellbore.isActive === "true"}
+      isActive={wellbore.isActive === "true" || wellbore.isActive === "1"}
     >
       <TreeItem
         nodeId={logGroupId}
--- a/Src/Witsml/Data/WitsmlWellbore.cs
+++ b/Src/Witsml/Data/WitsmlWellbore.cs
@@ -13,5 +13,7 @@ namespace Witsml.Data
         [XmlElement("purposeWellbore")] public string PurposeWellbore { get; set; }
         [XmlElement("typeWellbore")] public string TypeWellbore { get; set; }
         [XmlElement("commonData")] public WitsmlCommonData CommonData { get; set; }
+        [XmlElement("isActive")] public string IsActive { get; set; }
+
     }
 }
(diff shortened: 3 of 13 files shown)


Src/WitsmlExplorer.Frontend/components/Icons/IsActiveIcon.tsx
```diff
@@ -0,0 +1,9 @@
+﻿import * as React from "react";
+
+export function IsActiveIcon(): React.ReactElement {
+  return (
+    <>
```
steinsiv, line 5: should this be a div ? or is there some implicit conversion going on here in react.element?

Src/WitsmlExplorer.Frontend/components/Sidebar/WellboreItem.tsx
```diff
@@ -118,6 +118,7 @@ const WellboreItem = (props: WellboreItemProps): React.ReactElement => {
       labelText={wellbore.name}
       onLabelClick={onLabelClick}
       onIconClick={onIconClick}
+      isActive={wellbore.isActive === "true"}
```
steinsiv: Will this catch all string values from Store? e.g. "0" an "1" which is not prohibited as an xml boolean ref earlier note: 

> For Boolean items, the XML Schema standard supports use of both “true” and “1” and “false” and “0”. WITSML is consistent with XML, so also supports use of both.

steinsiv (commented):
Build, Test and Executed just fine at my end for all three projects Witsml, Api, Frontend.
`isActive=true` found on one wellbore on one WitsmlServer and icon shown as expected.

Src/WitsmlExplorer.Frontend/components/Sidebar/WellboreItem.tsx
HakonHesselberg: This will not catch all string values. I need to either compare 0 and 1 as well or turn it into a boolean somewhere.

Src/WitsmlExplorer.Frontend/components/Icons/IsActiveIcon.tsx
HakonHesselberg: This is a [React fragment](https://reactjs.org/docs/fragments.html).

In this case I don't think there is a need for either a div or a fragment, as there is only one root element anyways. If we have to choose one, <> is considered a [better choice](https://stackoverflow.com/questions/47761894/why-are-fragments-in-react-16-better-than-container-divs).

I see that some of the icons use <> and some use div. They should either all be the same or without a wrapper at all.

For this PR I will leave it as is, but I might clean up the rest laster by either removing the tag or replacing the div tags with <>.

marmid74:
I´ve tested against Equinor production witsml server and a BakerHuges witsmlserver (both type Sitecom) and verified that the isActive icon appears for the current active wells/wellbores.

I´ve tested against a Halliburton server (PDS witsml server) and verified that the isActive icon appears for active wells/wellbores.

marmid74:
Ran tests backend /frontend - all passed.

steinsiv (approved):

PR #332, opened 2021-04-12
Extract duplicate code into ModalParts and fix spelling in WellborePropertiesModal

HakonHesselberg:
## Description

Extract duplicate code from several modals into one class.
Fix spelling.

## Type of change

- [ ] Bugfix

* PropertiesModals

# Checklist:

- [ ] PR is related to an issue

### DIFF
--- a/Src/WitsmlExplorer.Frontend/components/Modals/ModalParts.tsx
+++ b/Src/WitsmlExplorer.Frontend/components/Modals/ModalParts.tsx
@@ -1,0 +1,8 @@
+﻿export enum PropertiesModalMode {
+  New,
+  Edit
+}
+
+export const validText = (text: string): boolean => {
+  return text && text.length > 0;
+};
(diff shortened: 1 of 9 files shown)


Src/WitsmlExplorer.Frontend/components/Modals/ModalParts.tsx
```diff
@@ -0,0 +1,8 @@
+﻿export enum PropertiesModalMode {
+  New,
+  Edit
+}
+
+export const validText = (text: string): boolean => {
+  return text && text.length > 0;
```
steinsiv, line 7:
```
>>  "" && "".length > 0;
<-   ""
```
An empty string in this expression returns string and not boolean, I have not testet through the application though. It might be handled implicitly there? All yarn tests ran fine

steinsiv (approved):
test run ok, comment below

Src/WitsmlExplorer.Frontend/components/Modals/ModalParts.tsx
marmid74: wrapping it like this should return a boolean
return !!(text && text.length> 0);

HakonHesselberg: I have not made any changes to the code here, just moved it into a separate class as it was duplicated several places.

HakonHesselberg: An empty string here will return false.

HakonHesselberg: https://stackoverflow.com/questions/11543735/why-the-result-of-booltrue-string-is-string-in-javascript

PR #338, opened 2021-04-15
backend changes for batch update of well properties

### LINKED ISSUE #303: 👉 Batch update well attributes - API changes
This task is related user story #179 

Add new job type and implementation for batch update of wells. Similar to ModifyWellJob job-type

Open for suggestions.

**Additional context**
Similar to ModifyWellJob job-type

WITSML Update Query
image

kdi-chandra:
Fixes #303

## Description
This pull request has changes/implementation of API for batch update of well properties 

## Type of change

- [ ] Bugfix
- [x] New feature (non-breaking change which adds functionality)

*  WitsmlExplorer.Api - ModifyWellJob, BatchModifyWellJob

# Checklist:

## Further comments
NA


Src/WitsmlExplorer.Api/Query/WellUpdateQueries.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.Collections.Generic;
+using Witsml.Data;
+using WitsmlExplorer.Api.Models;
+
+namespace WitsmlExplorer.Api.Query
+{
+    public static class WellUpdateQueries
```
HakonHesselberg: Should this code be in WellQueries.cs?

Src/WitsmlExplorer.Api/Query/WellUpdateQueries.cs
```diff
@@ -0,0 +1,34 @@
...
+            {
+                Wells = new List<WitsmlWell>
+                {
+                    new WitsmlWell
+                    {
+                        Uid = well.Uid,
+                        Name = well.Name,
+                        Field = well.Field,
+                        Country = well.Country,
+                        Operator = well.Operator
```
HakonHesselberg: Is time zone missing?

Src/WitsmlExplorer.Api/Query/WellUpdateQueries.cs
kdi-chandra: WitsmlExplorer.Api.Models.Well is DTO which can't be referenced in Witsml project. Hence this class is in WitsmlExplorer.Api project.

marmid74:
Verified code and ran tests. All tests passed. 👍

Src/WitsmlExplorer.Api/Workers/ModifyWellWorker.cs
```diff
@@ -31,7 +32,7 @@ public async Task<(WorkerResult, RefreshAction)> Execute(ModifyWellJob job)
             Verify(job.Well);
             var wellUid = job.Well.Uid;
             var wellName = job.Well.Name;
-            var query = WellQueries.UpdateQuery(wellUid, wellName);
+            var query = WellUpdateQueries.UpdateQuery(job.Well);
```
steinsiv: Is it possible to use `WellReference` here? or simply the string id just to conform with the other workers 
Witsml Query specifics could then be left to the `Witsml` project

Src/WitsmlExplorer.Api/Query/WellUpdateQueries.cs
steinsiv: If we use the well reference or id (string), we would avoid having Witsml Query specifics in this project?

Src/WitsmlExplorer.Api/Jobs/BatchModifyWellJob.cs
```diff
@@ -0,0 +1,9 @@
+using WitsmlExplorer.Api.Models;
+
+namespace WitsmlExplorer.Api.Jobs
+{
+    public class BatchModifyWellJob
+    {
+        public Well[] Wells { get; set; }
```
steinsiv: Maybe we can use `WellReference` here? e.g. like in `DeleteLogObjects.cs` 

> `        public IEnumerable<LogReference> LogReferences { get; set; }`

kdi-chandra: I see, But along with reference-id's we need to pass other parameters like (name, country, timezone, field etc...) for multiple wells. So it would be array of Well DTO's

Src/WitsmlExplorer.Api/Query/WellUpdateQueries.cs
kdi-chandra: To update all the properties of the well, we might need to pass all the properties of Well as parameters for WellQueries.UpdateQuery which will like this

WellQueries.UpdateQuery(wellId, wellName, field, country, timezone, operator, ...)

Thought better approach is pass Well DTO and create query.

Src/WitsmlExplorer.Api/Workers/ModifyWellWorker.cs
kdi-chandra: To update all the properties of the well, we might need to pass all the properties of Well as parameters for WellQueries.UpdateQuery which will like this

WellQueries.UpdateQuery(wellId, wellName, field, country, timezone, operator, ...)

Thought better approach is pass Well DTO and create query.

Src/WitsmlExplorer.Api/Jobs/BatchModifyWellJob.cs
HakonHesselberg: I think it's correct to use Well here as we are passing multiple parameters, like Chandra said. The same approach is used in for instance `ModifyWellJob.cs` and `CreateWellJob.cs`.

Src/WitsmlExplorer.Api/Query/WellUpdateQueries.cs
HakonHesselberg: I have reviewed more closely now and I agree with Chandra's implementation. Previously there were only two string parameters passed, which didn't have a need for a separate DTO, but now we are passing a full Well object.

I think in the future a similar approach will be necessary for the other workers as, if we want to modify more parameters than what are currently supported.

Src/WitsmlExplorer.Api/Workers/AbstractModifyWellWorker.cs
```diff
@@ -0,0 +1,25 @@
+using Witsml.Data;
+using Witsml.Extensions;
+using WitsmlExplorer.Api.Models;
+
+namespace WitsmlExplorer.Api.Workers
+{
+    public abstract class AbstractModifyWellWorker
```
HakonHesselberg: As there is nothing actually being inherited I prefer this as a static class how it was before.

kdi-chandra: Agree, I will improve this code in next work item

steinsiv: In my opinion this obscures the Query part and deviates further from the established codebase. 
But as commented before, I think it would be a cleaner solution not producing queries from both `Witsml` and `Witsml.Api` . What do you think?

kdi-chandra: Agree to some extent, But for 'Update' queries it has to be constructed in Witsml.Api in order to maintain cleaner code.

I see similar approach in DeleteLogObjectsWorker.CreateRequest

kdi-chandra: So, Now removed the abstract class as @HakonHesselberg commented, and added static method similar to DeleteLogObjectsWorker.

Will think of further optimization/coding practices as we progress

Src/WitsmlExplorer.Api/Query/WellUpdateQueries.cs
kdi-chandra: For now following same approach as DeleteLogObjectsWorker. Will improve further when needed as we progress

Src/WitsmlExplorer.Api/Workers/ModifyWellWorker.cs
kdi-chandra: For now following same approach as DeleteLogObjectsWorker. Will improve further when needed as we progress

Issue #354, opened 2021-04-26
📘 Document on how to deploy WE in any web server


kdi-chandra:
**Documentation Is:📘 **

- [x] Missing
- [x] Needed
- [ ] Confusing
- [ ] Not Sure?

**Please Explain in Detail**

Need to document all the steps required to deploy WE on IIS or any web server

marmid74:
Please submit a pull request with the improved documentation as suggested.

PR #364, opened 2021-04-30
Enable environment variables to support deployment on distributed sys…

### LINKED ISSUE #355: 🐛 WE Frontend and Backend can't be deployed on different hosts/paths
**:bug: Describe the bug**
We can't deploy WE frontend and backend on different domains/sub-domains

**To Reproduce**
Frontend code is expecting the API to be on the same host

**Expected behavior**
WE Frontend should be configurable to accept the WE-API's host/path

**Additional context**
ApiClient.getBaseUrl should return configured URL for the backend API

kdi-chandra:
Fixes #355 

## Description
This change set will enable us to deploy WE solution in secondary domains or on different host/path.
Added environment variables
- NEXT_PUBLIC_WE_URL -> specify the host/path for the WE Frontend deployment. Eg.: https://localhost/witsmlexplorer
- NEXT_PUBLIC_WE_API_URL -> specify the host/path for the WE Backend deployment. Eg.: https://localhost/witsmlexplorer.api
These environment variables are used in app at appropriate places to resolve all URL's of APIs, Icons, Fonts and base-href in index.html

## Type of change

- [x] Bugfix
- [x] This change requires a documentation update

* Application/Data loading
* Icons/Fonts rendering

# Checklist:

## Further comments
I have not updated the documentation, as we have issue #354 to add related content


marmid74:
I see that you indicate that this PR also requires an update of documentation, please include updated documentation as well as part of PR.

kdi-chandra:
Yes, it requires a new section in documentation, which is missing indicated by issue #354 . It can be done as part of issue #354 .

Src/WitsmlExplorer.Frontend/.env.local
```diff
@@ -0,0 +1,2 @@
+NEXT_PUBLIC_WE_URL=https://localhost/witsmlexplorer
+NEXT_PUBLIC_WE_API_URL=https://localhost/witsmlexplorer.api
```
steinsiv: I think we need to do something with the  `NEXT_PUBLIC_WE_API_URL` before any merging, to be able to use the backend without modification. Another alternative is to do the address change in both backend and frontend at the same time.

The frontend prepends `api` on the call so this would double up like `witsmlexplorer.api/api/`

Should be consider versioning the api? Tell me your thoughts on that

aszalos-roland (approved):
looks good to me 👍

Src/WitsmlExplorer.Frontend/.env.local
kdi-chandra: With this PR, backend is not modified, backend can run on any host/port/path. Eg. https://localhost:5555;
Then we set NEXT_PUBLIC_WE_API_URL =  https://localhost:5555

When the frontend prepends 'api' then it would be like https://localhost:5555/api/witsml-servers

Versioning API won't change this implementation, as we are parameterizing only host/port/path. When we introduce versioning of API then URL will look like  https://localhost:5555/api/v1/witsml-servers, which needs to be handled in backend/frontend code

marmid74: Awaits merge of this PR until documentation is provided how to setup local environment due this change by an pr on issue #354.

kdi-chandra: All the steps are documented and created a PR https://github.com/equinor/witsml-explorer/pull/385

Src/WitsmlExplorer.Frontend/components/AssetsLoader.ts
```diff
@@ -0,0 +1,6 @@
+export class AssetsLoader {
+  static getAssetsRoot() {
```
lindsve: Any warnings added by github should be fixed. This one is an easy fix, just add `: string` as return type

Src/WitsmlExplorer.Frontend/.env.local
lindsve: Using nextjs as the framework for the frontend application should remain as an implementation detail, and I don't see the need to expose that in the environment variable name. Could these be `WE_FRONTEND_URL` and `WE_API_URL` instead? (or use full name like `WITSMLEXPLORER_FRONTEND_URL` / `WITSMLEXPLORER_API_URL`)

kdi-chandra: Incorporated the review comments and updated the related documentation

Src/WitsmlExplorer.Frontend/components/AssetsLoader.ts
kdi-chandra: Taken care with latest commit

lindsve (approved):

steinsiv (approved):
Just a comment: `.env.local` should probably be added to gitignore, and maybe make use of `.env.development` and `.env.production` checked in

PR #376, opened 2021-05-10
When the user tries to open more than 200 mnemonics the request fails

### LINKED ISSUE #365: 👉 When the user tries to open more than 200 mnemonics the request fails
We allow the users to select all the mnemonics, but the get request fails when the URL is more than 2048 characters.
Related stories: #173, any story that relies on reading LogData

Write an alternative POST endpoint, so we can send the mnemonics in the request body if the URL would be too long.

Not allowing the user to select that many mnemonics.

**Additional context**
Screenshot of the problem:
too_many_mnemonics

aszalos-roland:
Fixes #365

## Description
I created a GetLargeLogData Post endpoint in the API, on the client side if the URI is too long it will use this endpoint otherwise the already existing one.

## Type of change
- [x] Bugfix

CurveValuesView, logObjectService

# Checklist:

## Further comments
Do we need to keep the original endpoint and make a request like the current implementation, or can we remove the old endpoint and use only the Post implementation?


kdi-chandra (approved):
Changes looks good

steinsiv (commented):
Works fine here, I left a comment on the GET/POST switch, maybe should use POST in all cases to simplify the codebase? What do you think?
There was a conflict in one file, when the conflict is cleared we can merge this 👍

HakonHesselberg (approved):
Tested and works locally. Looks good to me 🍪

HakonHesselberg:
> Works fine here, I left a comment on the GET/POST switch, maybe should use POST in all cases to simplify the codebase? What do you think?

I think this should be considered, but it is out of scope for this PR. I think the PR should be merged once the conflict is solved and then we can make a separate issue for the get/post if necessary.

PR #453, opened 2021-06-09
Add missing COPY for Console.csproj in Dockerfile-api

HakonHesselberg:
Thank you for contributing to WITSML Explorer!
Before submitting this PR, please fill in the following template.

## Fixes
This pull request does not relate to any issue.

build_docker_images.sh currently fails because it is expecting to find Console.csproj, which is not copied.

## Description

This PR adds missing COPY statement of Console.csproj in the Docker-api file.

## Type of change
_List the types of change. Remove from the list any points which are not necessary_

* Bugfix

* Docker builds

## Checklist:
_Please tick all the boxes or remove the ones that aren't needed and explain why_

*Code quality*

*Test coverage*
* [x] Existing tests pass


lindsve (approved):
A quickfix that solves the issue, but makes it apparent that the dockerfiles need to be revised to not depend on the solution file. This one already has a dependency it shouldn't have (frontend), so this should be fixed

PR #481, opened 2021-06-17
Added list risks command to console app

lindsve:
This pull requests adds a list risks command to the console application.

It currently has the following options:
```
USAGE:
    dotnet run -- list risks [OPTIONS]

OPTIONS:
    -h, --help           Prints help information
        --uidWell        Well UID
        --uidWellbore    Wellbore UID
        --source         Source for the risk
        --lastChanged    A valid timestamp for when the risk was last changed (example: 2021-06-17T01:37:16.594Z)
```

To minimize the potential result set, it is required to filter on either uidWell/uidWellbore or lastChanged. It will also give an error if the result set is larger than 100 risks (since this won't be that usable in a cli environment).

## Type of change
* New feature (non-breaking change which adds functionality)

## Impacted Areas in Application
* Console application

## Checklist:
*Communication*
* [ ] PR is related to an issue
* [ ] I have made corresponding changes to the documentation

*Code quality*
* [x] Code follows the style guidelines
* [x] I have self-reviewed my code
* [x] No new warnings are generated

*Test coverage*
* [x] Existing tests pass
* [ ] New code is covered by passing tests

PR #486, opened 2021-06-17
Console application: Extract settings to separate files

lindsve:
@HakonHesselberg commented on #481 if we should have public classes nested inside other public classes. I don't have any strong opinions on this, but this is a PR with a proposal for a change that moves the settings files to individual files/classes instead of being nested inside the commands


HakonHesselberg:
I don't mind the classes being nested if they are private, but I think best practice for public classes is to keep them in separate files.

stigrus (approved):

HakonHesselberg (approved):

PR #490, opened 2021-06-18
Fix missing Get statement for GetLogsForWellbore

HakonHesselberg:
Fixes missing Get Statement for GetLogsForWellbore


steinsiv (approved):
Thanks! 😊

Issue #174, opened 2021-02-16
User story:  Import  log from .csv


thomash1972:
As a Data Manager
I want to be able to Import log data from a defined .csv file format to a WITSML server in the context of a selected wellbore.

### Acceptance criteria ###
- on an existing log object , import additional log data from a pre-defined .csv format (same format as Export log user story.)
- If the import .csv format is not in the correct format, WE will inform me about this without trying to import the data

Also, here are a few points that need to be addressed:
- [X] The target Log need to exist before importing the data from right clicking on the Log Object
- [ ]  alternatively the import function will create the Log if needed (Log name then need to be specified in the .csv file ? 
- [ ] Alternatively you could allow the import to be initiated from a generic menu item, but require that the well, wellbore as specified in the .csv must exist, and prompt the user for which existing Log the data should be imported to
- [X] If a curve does not exists in log, the solution must resolve the datatype upfront. This can be done based on unit. If unitless, the solution must resolve if the datatype is string, numeric or date
- [X] If a curve does not exists, solution must add values on mandatory properties on curve (datatype, unit, uid?, description?).
- [X] Witsml servers Datapoint limitation - maximum 10 000 rows for each updateInStore, so .csv with more rows, need to be divided into multiple updateInStore operations.

### Resources ###
-  Development: TDE
-  Lead: Equinor

HoldampfIldiko:
DEV- TDE @aszalos-roland

marmid74:
This user story is named VERSION 1, of Import log from .csv. Implemented by TDE Group.
Extended features, named VERSION 2, of Import log from .csv is specified in user story #554

PR #494, opened 2021-06-21
#174 import log from .csv

aszalos-roland:
## Fixes
This pull request is an implementation of the story #174

## Description
Added new import data option in the LogObject context menu as seen in the following image:
image

Clicking this option brings up a modal with the option to upload a .csv file to upload data to the selected log as seen in the following image:
image

The uploaded file is validated against the existing columns in the chosen log as seen in the following image:
image

The implementation satisfies the acceptance criteria described in the story #174, however there can be improvements made.
These improvements are mentioned as Todo's in the code.

## Type of change
* This change requires a documentation update

Backend:
* Workers
* Tests

Frontend:
* CurveValuesView
* LogObjectContextMenu
* Changed some icon sizes so they match in the context menu

## Checklist:
*Communication*
* [x] PR is related to a story

*Code quality*

*Test coverage*
* [x] Existing tests pass

## Further comments
Changed the export to always sort by Time or Depth when exporting, this way when we import the exported data it doesn't fail.


steinsiv (commented):
I think this will be very helpful contribution! 👍 Tests run, and import working fine locally on Witsml test server.

Some small comments
1. empty newlines at end corrupt import
2. mnemonics must exist (no create if missing/empty log)
3. existing values with same index will overwrite w/o warnings

Number 2. and 3. might be as expected by design, just thought I should mention it. 
Number 1. we could prevent by sanitizing trailing newlines

aszalos-roland:
> I think this will be very helpful contribution! 👍 Tests run, and import working fine locally on Witsml test server.
> 
> Some small comments
> 
> 1. empty newlines at end corrupt import
> 2. mnemonics must exist (no create if missing/empty log)
> 3. existing values with same index will overwrite w/o warnings
> 
> Number 2. and 3. might be as expected by design, just thought I should mention it.
> Number 1. we could prevent by sanitizing trailing newlines

Thanks for the review.
Yea Number 1 can be easily implemented, however I think it's optional because the target format comes from the Export story.
Number 3 was not discussed in the story, but we could implement a checkbox for the user to check if they want to overwrite existing records and then we can only send the new records to the server for storage or something like that.
Number 2 might need further discussions if we want this feature.

steinsiv (approved):

PR #575, opened 2021-07-22
Modified the Dockerfile for the api so that it is no longer solution dependant

### LINKED ISSUE #454: Improve docker files
The current dockerfiles (f.ex Dockerfile-api) builds based on the solution file. This makes it dependant on all projects in the solution, although it should only require the projects for that particular dockerfile. In the case of Dockerfile-api, it depends on the frontend project. This should not be included (and also the newly added console project).

The solution is to build based on the project files themselves instead of the solution file

lindsve:
## Fixes
This pull request fixes #454 

## Description
This changes the Dockerfile for the api project so that it is no longer dependant on the solution file, and instead builds and tests the `Witsml` and `WitsmlExplorer.Api` projects explicitely.

This makes it no longer dependant on copying every project file to the docker context before building inside the docker image.

...

## Type of change

* Dockerfile-Api

## Checklist:

*Communication*

*Code quality*

*Test coverage*
* [ ] Existing tests pass

### DIFF
--- a/Dockerfile-api
+++ b/Dockerfile-api
@@ -1,22 +1,21 @@
 FROM mcr.microsoft.com/dotnet/sdk:5.0 as base
 WORKDIR /app
-COPY witsml-explorer.sln .
 COPY Src/Witsml/Witsml.csproj Src/Witsml/Witsml.csproj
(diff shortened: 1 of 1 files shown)


steinsiv (approved):
Tested locally, built api image and ran in concert with frontend and mongo-db via docker-compose. all ok 👍

PR #1338, opened 2022-07-13
WE-490 Replace Carter

steinsiv:
## Fixes
This pull request fixes #1107 

## Description
Carter v5 is outdated and needs to be replaced with minimal WebApi

See Issue: https://github.com/equinor/witsml-explorer/issues/1107 
and closed PR: https://github.com/equinor/witsml-explorer/pull/1106

## Type of change
* Enhancement of existing functionality

## Impacted Areas in Application
* API

## Checklist:

*Communication*
* [x] PR is related to an issue
* [ ] I have made corresponding changes to the documentation

*Code quality*
* [x] Code follows the style guidelines
* [x] I have self-reviewed my code
* [x] No new warnings are generated

*Test coverage*
* [x] Existing tests pass
* [ ] New code is covered by passing tests

PR #1414, opened 2022-08-16
WE-505 swagger

steinsiv:
## Further comments
This is a follow up for https://github.com/equinor/witsml-explorer/pull/1338 with swagger UI generation

Needs additional settings in `mysettings.json` for the Oauth2 part to work

```json
{
    "AzureAd": {
        "AppName": "<appname>",
        "Instance": "https://login.microsoftonline.com/",
        "Domain": "qualified.domain.name",
        "TenantId": "<tenant>",
        "ClientId": "<client>",
        "Scopes": "User.Read <client>/access_as_user",
        "TokenValidationParameters": {
            "ValidateAudience": false
        },
        "Swagger": {
            "AuthorizationUrl": "https://login.microsoftonline.com/<tenant>/oauth2/v2.0/authorize",
            "TokenUrl": "https://login.microsoftonline.com/<tenant>/oauth2/v2.0/token"
        }
    }
}
````

### DIFF
--- a/Src/WitsmlExplorer.Api/Routes.cs
+++ b/Src/WitsmlExplorer.Api/Routes.cs
@@ -12,47 +12,47 @@ public static class Api
 {
     public static void ConfigureApi(this WebApplication app)
     {
-        app.MapGet("/api/witsml-servers", WitsmlServerHandler.GetWitsmlServers).Produces<IEnumerable<Server>>();
-        app.MapPost("/api/witsml-servers", WitsmlServerHandler.CreateWitsmlServer).Produces<Server>();
-        app.MapMethods("/api/witsml-servers/{witsmlServerId}", new[] { HttpMethods.Patch }, WitsmlServerHandler.UpdateWitsmlServer);
-        app.MapDelete("/api/witsml-servers/{witsmlServerId}", WitsmlServerHandler.DeleteWitsmlServer).Produces(StatusCodes.Status204NoContent);
+        app.MapGet("/api/witsml-servers", WitsmlServerHandler.GetWitsmlServers).Produces<IEnumerable<Server>>().Produces(StatusCodes.Status500InternalServerError);
+        app.MapPost("/api/witsml-servers", WitsmlServerHandler.CreateWitsmlServer).Produces<Server>().Produces(StatusCodes.Status500InternalServerError);
+        app.MapMethods("/api/witsml-servers/{witsmlServerId}", new[] { HttpMethods.Patch }, WitsmlServerHandler.UpdateWitsmlServer).Produces(StatusCodes.Status500InternalServerError);
+        app.MapDelete("/api/witsml-servers/{witsmlServerId}", WitsmlServerHandler.DeleteWitsmlServer).Produces(StatusCodes.Status204NoContent).Produces(StatusCodes.Status500InternalServerError);
 
-        app.MapGet("/api/wells", WellHandler.GetAllWells).Produces<IEnumerable<Well>>();
-        app.MapGet("/api/wells/{wellUid}", WellHandler.GetWell).Produces<Well>();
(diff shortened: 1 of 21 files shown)

### CODE REVIEW

File: Src/WitsmlExplorer.Api/Routes.cs [hunk 1]
```diff
@@ -1,55 +1,63 @@
...
 
 public static class Api
 {
     public static void ConfigureApi(this WebApplication app)
     {
-        app.MapGet("/api/witsml-servers", WitsmlServerHandler.GetWitsmlServers);
-        app.MapPost("/api/witsml-servers", WitsmlServerHandler.CreateWitsmlServer);
-        app.MapMethods("/api/witsml-servers/{witsmlServerId}", new[] { HttpMethods.Patch }, WitsmlServerHandler.UpdateWitsmlServer);
-        app.MapDelete("/api/witsml-servers/{witsmlServerId}", WitsmlServerHandler.DeleteWitsmlServer);
+        app.MapGet("/api/witsml-servers", WitsmlServerHandler.GetWitsmlServers).Produces<IEnumerable<Server>>().Produces(StatusCodes.Status500InternalServerError);
```
lindsve: I propose that you add some linebreaks to improve the readability a bit here:
```
app.MapGet("/api/witsml-servers", WitsmlServerHandler.GetWitsmlServers)
   .Produces<IEnumerable<Server>>()
   .Produces(StatusCodes.Status500InternalServerError);
```

You could also take a look at `builder.Services.AddEndpointsApiExplorer()`, which should annotate endpoints automatically without these `.Produces` statements (although I suspect it might not correctly annotate error codes)

lindsve: Seems like we were already using `AddEndspointsApiExplorer` and it might also be required 🤔 

It also does not include the model description for the endpoint results, so there are trade-offs that can be made here. I like that the OpenAPI UI describes the result models, but I'm not a huge fan of adding all these `.Produces` calls 🤷

lindsve: Another alternative to adding all these `.Produces` calls in the route configurations is to use the Produces attribute on the handler methods as well:
```
[Produces(typeof(IEnumerable<Server>))]
```
This is closer to the actual implementation

steinsiv: Yes that is a good idea, it is close to the handler, so that makes sense.
We usually only return 200 statuscode and 500. I guess they can be attributed as well 👍

Src/WitsmlExplorer.Api/Startup.cs
```diff
@@ -57,6 +61,48 @@ public void ConfigureServices(IServiceCollection services)
             services.ConfigureDependencies(Configuration);
             services.AddHostedService<BackgroundWorkerService>();
             services.AddScoped<ICopyLogDataWorker, CopyLogDataWorker>();
+            services.AddEndpointsApiExplorer();
+            services.AddSwaggerGen(options =>
+            {
+                var basicSecurityScheme = new OpenApiSecurityScheme
```
janburak: Maybe move out this code block to its own file a la "SetupSwagger" to keep Startup more concise?

Src/WitsmlExplorer.Api/Startup.cs
```diff
@@ -76,6 +132,9 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
             app.UseResponseCompression();
             app.UseCors(_myAllowSpecificOrigins);
             app.UseRouting();
+            //app.UseHttpsRedirection();
```
janburak: Leftover comment?

Src/WitsmlExplorer.Api/Startup.cs
steinsiv: Yes that will be cleaner. I will fix 👍

janburak (approved):
- [X] Read issue and PR
- [X] Reviewed the code

PR #1602, opened 2022-10-10
WE-582 increase number of roles on servers

steinsiv:
## Fixes
This pull request fixes WE-582 + WE-577

## Description
* Expand on `Server`s in configuration database, to be able to add more roles to a server allowing for better access control of who gets to use system credentials for servers under the `OAuth2`securityscheme

* Add tests to `CredentialsService` public `GetCreds` method

* Rewrite `WitsmlClientProvider` to reduce load on init. Pass header values to `Credentialsservice` as an alternative.

* Move responsibility for getting `WitsmlClients` to `Baseworker`

## Type of change

* `Breaking change` (fix or feature that would cause existing functionality to not work as expected)

## 💥 Potential breaking change!
Server objects in config database (cosmosdb or mongodb) needs to be updated to fit the new structure as outlined in `swagger` spec under `https://localhost:5001/swagger` for endpoint `api/witsml-servers`. Any properties not in the list below needs to be changed/removed.

```json
{
  "id": "3fa85f64-5717-4562-b3fc-2c963f66afa6",
  "name": "string",
  "url": "string",
  "description": "string",
  "securityscheme": "string",
  "roles": [
    "string"
  ]
}
```

- Config database (Servers)
- API (Server model object)
- Frontend (Server manager)

## Checklist:

*Communication*

*Code quality*

*Test coverage*
* [x] Existing tests pass

## Further comments
Good improvements, but `WitsmlClientProvider` is still being initialized for every worker despite only 1 job is executed pr scope. This should be looked into.


Src/WitsmlExplorer.Api/Services/CredentialsService.cs
```diff
@@ -86,7 +86,7 @@ private async Task<bool> UserHasRoleForHost(string[] roles, Uri host)
 
             bool systemCredsExists = _witsmlServerCredentials.WitsmlCreds.Any(n => n.Host == host);
             IEnumerable<Server> hostServer = allServers.Where(n => n.Url.ToString() == host.ToString());
-            bool validRole = hostServer.Any(n => roles.Contains(n.Role));
+            bool validRole = hostServer.Any(n => n.Roles.Intersect(roles).Any());
```
janburak, line 97: Would Server.Roles ever be empty to allow anybody to access it? In that case, this line would fail.

janburak:
There seems to be a prior omission regarding setting credentials in CredentialsService. With Basic, the flow seems to be
1. ServerManager fetches the servers and saves them in the navigation state.
2. A user picks a server from the list.
3. The user enters credentials for the picked server.
4. CredentialsService saves the credentials both in memory and in local storage.

With OAuth, step 3 is omitted, and thus updateLocallyStoredServerUsernames is never run in step 4 (why are they saved in local storage when we fetch the servers anyway?). This means that ApiClient will not set server headers because no credentials are initialized in CredentialsService.
A solution to this would be for CredentialsService to update the credentials list based on the api/witsml-servers route.
Edit: fixed by https://github.com/equinor/witsml-explorer/pull/1606

Src/WitsmlExplorer.Api/Services/CredentialsService.cs
steinsiv: Yes I think it will be empty in some cases but we should probably not grant access based on missing or empty role assignments. When roles are missing or not configured the default will be no access until given explicitly

steinsiv:
> There seems to be a prior omission regarding setting credentials in CredentialsService. With Basic, the flow seems to be
> 
> 1. ServerManager fetches the servers and saves them in the navigation state.
> 2. A user picks a server from the list.
> 3. The user enters credentials for the picked server.
> 4. CredentialsService saves the credentials both in memory and in local storage.
> 
> With OAuth, step 3 is omitted, and thus updateLocallyStoredServerUsernames is never run in step 4 (why are they saved in local storage when we fetch the servers anyway?). This means that ApiClient will not set server headers because no credentials are initialised in CredentialsService. A solution to this would be for CredentialsService to update the credentials list based on the api/witsml-servers route.

I agree and think the whole flow might need a little revamp now. Localstorage stores `username:URL`, and in-memory the password for the same is stored, so it seems redundant. Also not providing a Header for `WitsmlTargetServer` with either an `URL` or `b64(creds):URL` in step 4 would be a bug that needs to be fixed. Should we assign this to a separate PR?

janburak:
> There seems to be a prior omission regarding setting credentials in CredentialsService. With Basic, the flow seems to be
> 
> 1. ServerManager fetches the servers and saves them in the navigation state.
> 2. A user picks a server from the list.
> 3. The user enters credentials for the picked server.
> 4. CredentialsService saves the credentials both in memory and in local storage.
> 
> With OAuth, step 3 is omitted, and thus updateLocallyStoredServerUsernames is never run in step 4 (why are they saved in local storage when we fetch the servers anyway?). This means that ApiClient will not set server headers because no credentials are initialized in CredentialsService. A solution to this would be for CredentialsService to update the credentials list based on the api/witsml-servers route.

janburak (approved):
- [X] Read issue and PR
- [X] Reviewed the code

PR #1603, opened 2022-10-11
WE-577 ensure source server is set on copyjobs

steinsiv:
## Fixes
This pull request fixes WE-577

## Description
Workers initiating copy jobs would fallback on the same server as destination if source server request `Header` named `WitsmlSourceServer` had not been set.

## Type of change

:bug: Bugfix

`API` Workers

*Communication*

*Code quality*

*Test coverage*
* [x] Existing tests pass


janburak:
Job ordering fails on jobs other than copy, due to all workers being initialized when ordering a job (jobs other than copy do not specify a source server). Related to WE-584.

steinsiv:
> Job ordering fails on jobs other than copy, due to all workers being initialized when ordering a job (jobs other than copy do not specify a source server). Related to WE-584.

Yes, good point. We can fix that and join those two issues in the same PR? It feels strange that we need to initialise several copy workers when we are just deleting something (deleting does not need a source server reference)

steinsiv:
Closing due to changes incorporated in WE-582

PR #1625, opened 2022-10-26
WE-613 use cookies over in-memory handling

steinsiv:
## Fixes
This pull request fixes `WE-613` and `WE-612`

## Description
Remove handling of passwords from frontend and use cookies exclusively as a replacement for Basic auth against WITSML servers.

## Type of change

API, Frontend

## Checklist:

*Communication*

*Code quality*

*Test coverage*
* [x] Existing tests pass

## Further comments
Is part of story WE-586 and will be followed up by `WE-614` and `WE-615` to reduce header sizes for Basic auth, and support OAUTH along with no cookies and no frontend credentials handling.

### DIFF
--- a/Src/WitsmlExplorer.Api/HttpHandlers/AuthorizeHandler.cs
+++ b/Src/WitsmlExplorer.Api/HttpHandlers/AuthorizeHandler.cs
@@ -35,40 +35,6 @@ namespace WitsmlExplorer.Api.HttpHandlers
                 return Results.Ok();
             }
         }
-
-        // public static async Task<IResult> AuthorizeAndSetCookie([FromServices] ICredentialsService credentialsService, IHttpContextAccessor httpContextAccessor)
-        // {
-        //     string basicAuth = httpContextAccessor?.HttpContext?.Request.Headers[WitsmlClientProvider.WitsmlTargetServerHeader];
-        //     ServerCredentials creds = await credentialsService.GetCredentialsFromHeaderValue(basicAuth);
-        //     string encryptedPassword = await credentialsService.ProtectBasicAuthorization(basicAuth);
-
-        //     CookieOptions cookieOptions = new()
-        //     {
-        //         SameSite = SameSiteMode.Strict,
-        //         MaxAge = TimeSpan.FromDays(1),
-        //         Secure = true,
-        //         HttpOnly = true
-        //     };
-        //     string cookieValue = Convert.ToBase64String(Encoding.UTF8.GetBytes($"{creds.UserId}:{encryptedPassword}"));
-        //     httpContextAccessor.HttpContext.Response.Cookies.Append(Uri.EscapeDataString(creds.Host.ToString()), cookieValue, cookieOptions);
-        //     return Results.Ok(encryptedPassword);
-        // }
-
-        // public static IResult AuthorizeWithCookie([FromServices] ICredentialsService credentialsService, IHttpContextAccessor httpContextAccessor)
-        // {
-        //     ServerCredentials targetServer = httpContextAccessor?.HttpContext?.Request.GetWitsmlServerHttpHeader(WitsmlClientProvider.WitsmlTargetServerHeader, n => "");
-        //     IRequestCookieCollection cookies = httpContextAccessor.HttpContext.Request.Cookies;
-        //     if (targetServer.Host == null || !cookies.TryGetValue(Uri.EscapeDataString(targetServer.Host.ToString()), out string cookie))
-        //     {
-        //         return Results.Unauthorized();
-        //     }
-        //     if (!credentialsService.ValidEncryptedBasicCredentials($"{cookie}@{targetServer.Host}"))
-        //     {
-        //         return Results.Unauthorized();
-        //     }
-        //     return Results.Ok(cookie);
-        // }
-
         public static IResult Deauthorize(IHttpContextAccessor httpContextAccessor)
         {
             foreach (KeyValuePair<string, string> cookie in httpContextAccessor.HttpContext.Request.Cookies)
--- a/Src/WitsmlExplorer.Api/HttpHandlers/AuthorizeHandler.cs
+++ b/Src/WitsmlExplorer.Api/HttpHandlers/AuthorizeHandler.cs
@@ -1,57 +1,73 @@
 using System;
 using System.Collections.Generic;
-using System.Text;
 using System.Threading.Tasks;
 
 using Microsoft.AspNetCore.Http;
 using Microsoft.AspNetCore.Mvc;
 
 using WitsmlExplorer.Api.Configuration;
-using WitsmlExplorer.Api.Extensions;
 using WitsmlExplorer.Api.Services;
 
 namespace WitsmlExplorer.Api.HttpHandlers
 {
     public static class AuthorizeHandler
     {
-        public static async Task<IResult> Authorize([FromServices] ICredentialsService credentialsService, HttpRequest httpRequest)
+        public static async Task<IResult> Authorize([FromQuery(Name = "keep")] bool keep, [FromServices] ICredentialsService credentialsService, HttpContext httpContext)
         {
-            string basicAuth = httpRequest.Headers[WitsmlClientProvider.WitsmlTargetServerHeader];
-            return Results.Ok(await credentialsService.ProtectBasicAuthorization(basicAuth));
-        }
-
-        public static async Task<IResult> AuthorizeAndSetCookie([FromServices] ICredentialsService credentialsService, IHttpContextAccessor httpContextAccessor)
-        {
-            string basicAuth = httpContextAccessor?.HttpContext?.Request.Headers[WitsmlClientProvider.WitsmlTargetServerHeader];
+            string basicAuth = httpContext?.Request.Headers[WitsmlClientProvider.WitsmlTargetServerHeader].ToString();
             ServerCredentials creds = await credentialsService.GetCredentialsFromHeaderValue(basicAuth);
-            string encryptedPassword = await credentialsService.ProtectBasicAuthorization(basicAuth);
 
-            CookieOptions cookieOptions = new()
-            {
-                SameSite = SameSiteMode.Strict,
-                MaxAge = TimeSpan.FromDays(1),
-                Secure = true,
-                HttpOnly = true
-            };
-            string cookieValue = Convert.ToBase64String(Encoding.UTF8.GetBytes($"{creds.UserId}:{encryptedPassword}"));
-            httpContextAccessor.HttpContext.Response.Cookies.Append(Uri.EscapeDataString(creds.Host.ToString()), cookieValue, cookieOptions);
-            return Results.Ok(encryptedPassword);
-        }
-
-        public static IResult AuthorizeWithCookie([FromServices] ICredentialsService credentialsService, IHttpContextAccessor httpContextAccessor)
-        {
-            ServerCredentials targetServer = httpContextAccessor?.HttpContext?.Request.GetWitsmlServerHttpHeader(WitsmlClientProvider.WitsmlTargetServerHeader, n => "");
-            IRequestCookieCollection cookies = httpContextAccessor.HttpContext.Request.Cookies;
-            if (targetServer.Host == null || !cookies.TryGetValue(Uri.EscapeDataString(targetServer.Host.ToString()), out string cookie))
+            if (creds.IsCredsNullOrEmpty())
             {
                 return Results.Unauthorized();
             }
-            if (!credentialsService.ValidEncryptedBasicCredentials($"{cookie}@{targetServer.Host}"))
+            else
             {
-                return Results.Unauthorized();
+                string encryptedCredentials = await credentialsService.ProtectBasicAuthorization(basicAuth);
+                CookieOptions cookieOptions = new()
+                {
+                    SameSite = SameSiteMode.Strict,
+                    MaxAge = keep ? TimeSpan.FromDays(1) : null,
+                    Secure = true,
+                    HttpOnly = true
+                };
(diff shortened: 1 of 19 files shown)


Src/WitsmlExplorer.Api/Services/CredentialsService.cs
```diff
@@ -51,7 +49,22 @@ public async Task<string> ProtectBasicAuthorization(string headerValue)
...
-            return Encrypt(credentials.Password);
+            return Encrypt($"{credentials.UserId}:{credentials.Password}");
+        }
+
+        public async Task<ServerCredentials> GetCredentialsCookieFirst(IEssentialHeaders headers, string server)
+        {
+            ServerCredentials result;
+            if (headers.HasCookieCredentials(server))
+            {
+                var cred = Decrypt(headers.GetCookie(server)).Split(":");
```
janburak: Pretty much a non-issue but I get "Use explicit type instead of 'var'" here. Autoformatting being sluggish?

Tests/WitsmlExplorer.Api.Tests/Services/CredentialsServiceTests.cs
```diff
@@ -168,6 +169,41 @@ public void GetUsernames_ValidTokenUserValidBasicUserValidRole_ReturnUsernames()
...
 
+        [Fact]
+        public void GetCredentialsCookieFirst_ValidCookie_ReturnCookieCreds()
+        {
+            string host = "http://some.targeturl.com/";
+            ServerCredentials scTarget = new() { UserId = "cookieuser", Password = "cookiepass", Host = new Uri(host) };
+            IEssentialHeaders essentialHeaders = CreateEssentialHeaders(scTarget, null);
+            ServerCredentials serverCreds = _credentialsService.GetCredentialsCookieFirst(essentialHeaders, EssentialHeaders.WitsmlTargetServer).Result;
+
+            Assert.True(serverCreds.UserId == "cookieuser" && serverCreds.Password == "cookiepass" && serverCreds.Host.ToString() == host);
```
janburak: Shouldn't each comparison be its own assert?

Src/WitsmlExplorer.Frontend/services/apiClient.tsx
```diff
@@ -38,11 +37,12 @@ export class ApiClient {
     pathName: string,
     abortSignal: AbortSignal | null = null,
     currentCredentials = CredentialsService.getCredentials(),
-    includeCredentials = false
+    includeCredentials = true,
```
janburak: includeCredentials is AFAIK always set to true and can thus be removed from the signature

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
```diff
@@ -78,76 +83,69 @@ class CredentialsService {
...
+
+  // refresh timestamp for all "session" entries in localstorage, to match session cookies
+  public refreshLocalstorageSessions() {
+    const expirationTime = new Date();
+    expirationTime.setHours(expirationTime.getHours() + 1);
+    const cookies: CookieInfo[] = this.allLocalStorage()
+      .map((n) => JSON.parse(n))
+      .filter((n) => n.type == "WExSession");
+    cookies.forEach((element) => {
+      localStorage.removeItem(element.url);
```
janburak: Unnecessary removeItem call

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
```diff
@@ -78,76 +83,69 @@ class CredentialsService {
...
 
   public hasValidCookieForServer(serverUrl: string): boolean {
     //use local storage to check whether the cookie is valid, because the cookie is httpOnly
-    const cookieTimestamp = localStorage.getItem(serverUrl);
-    if (!cookieTimestamp) {
+    const cookieInfo = localStorage.getItem(serverUrl);
+    if (!cookieInfo) {
       return false;
+    } else {
+      const cInfo: CookieInfo = JSON.parse(cookieInfo);
```
janburak, line 97: JSON.parse will crash the website on invalid data (for example if somebody used WEx Dev before this PR)

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
```diff
@@ -78,76 +83,69 @@ class CredentialsService {
...
-      const creds = decoded.split(":");
-      return {
-        server: credentials.server,
-        username: creds[0],
-        password: creds[1]
-      };
-    } else {
-      const { message }: ErrorDetails = await response.json();
-      CredentialsService.throwError(response.status, message);
+  private allLocalStorage() {
```
janburak, line 122: Why not just loop over credentials like in deauthorize()?

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
```diff
@@ -10,6 +10,11 @@ export interface BasicServerCredentials {
   password?: string;
 }
 
+type CookieInfo = {
+  type: "WExSession" | "WExPersistent" | "WExInvalid";
+  url: string;
```
janburak, line 15: Why is the url part of the value if it is already the key?

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
```diff
@@ -78,76 +83,69 @@ class CredentialsService {
...
+    if (!cookieInfo) {
       return false;
+    } else {
+      const cInfo: CookieInfo = JSON.parse(cookieInfo);
+      return new Date().getTime() < new Date(cInfo.expiry).getTime();
     }
-    return new Date().getTime() < new Date(cookieTimestamp).getTime();
-  }
-
-  public keepLoggedInToServer(serverUrl: string): boolean {
```
janburak, line 113: Should this functionality be reimplemented instead of removed? Could save the keepLoggedIn value in CookieInfo if so.

Src/WitsmlExplorer.Frontend/components/Modals/UserCredentialsModal.tsx
```diff
@@ -89,12 +53,15 @@ const UserCredentialsModal = (props: UserCredentialsModalProps): React.ReactElem
       password
     };
     try {
-      const encryptedPassword = await CredentialsService.verifyCredentials(credentials);
-      props.onConnectionVerified({ ...credentials, password: encryptedPassword });
```
janburak, line 93: Not calling props.onConnectionVerified breaks every use of CredentialsMode.TEST

Src/WitsmlExplorer.Api/HttpHandlers/AuthorizeHandler.cs
```diff
@@ -1,58 +1,40 @@
...
             }
-            if (!credentialsService.ValidEncryptedBasicCredentials($"{cookie}@{targetServer.Host}"))
+            else
             {
-                return Results.Unauthorized();
+                string encryptedCredentials = await credentialsService.ProtectBasicAuthorization(basicAuth);
+                CookieOptions cookieOptions = new()
+                {
+                    SameSite = SameSiteMode.Strict,
+                    MaxAge = keep ? TimeSpan.FromDays(1) : null,
```
janburak, line 30: "Session cookies are deleted when the current session ends. The browser defines when the "current session" ends, and some browsers use session restoring when restarting. **This can cause session cookies to last indefinitely**." 
https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies
This means that by not checking "keep me logged in", we actually set a cookie that last indefinitely in some cases. Maybe we need to rethink how we should use cookies with that in mind.

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
```diff
@@ -78,76 +83,69 @@ class CredentialsService {
...
   public isAuthorizedForServer(server: Server): boolean {
     if (msalEnabled && server?.securityscheme == SecurityScheme.OAuth2 && getUserAppRoles().some((x) => server.roles.includes(x))) {
       return true;
     }
-    return this.credentials.find((c) => c.server.id === server?.id)?.password !== undefined;
-  }
-
-  public hasPasswordForUrl(serverUrl: string) {
-    return this.credentials.find((c) => c.server.url === serverUrl)?.password !== undefined;
+    return this.hasValidCookieForServer(server?.url);
```
janburak, line 90: Having a valid cookie does not equal the previous functionality of having a password (which we are ensured that it came from the authorize route). This for example cause WellProgress to show when we are unable to connect to the server even though we have a valid cookie.

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
steinsiv: good catch, we should apply `isJsonString` on this one aswell

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
steinsiv: that might be a better approach, I will check that out

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
steinsiv: I tried to only use `setItem` but the item did not update unless I removed it first.

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
steinsiv: It was not trivial to access the key when fetching all relevant items in localstorage, see `refreshLocalstorageSessions` 
It might be irrelevant if using the credentials list to access localstorage so I will test that next.

Tests/WitsmlExplorer.Api.Tests/Services/CredentialsServiceTests.cs
steinsiv: It might ? I think that is up for debate. I was here following the "only one assert pr test" flavour. 
This is something I can split up if it improves readability

Src/WitsmlExplorer.Frontend/services/apiClient.tsx
steinsiv: true, i think also the `authorization`endpoint uses it now. I will remove it thanks!

Src/WitsmlExplorer.Frontend/components/Modals/UserCredentialsModal.tsx
steinsiv: ok, i'll reinstate it

Src/WitsmlExplorer.Api/Services/CredentialsService.cs
steinsiv: yes, did not get that here? but true, it should,  I will replace it

Src/WitsmlExplorer.Api/HttpHandlers/AuthorizeHandler.cs
steinsiv: Ok, let's address that :)

Tests/WitsmlExplorer.Api.Tests/Services/CredentialsServiceTests.cs
janburak: I see your point. I am more used to something like this https://softwareengineering.stackexchange.com/a/171502
I guess if the contents of the assert are abstracted as "boolean receivedValidCookie" then a single assert should hold.

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
janburak: Hmm, on my end the item updates without removing it first.

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
steinsiv: yes maybe we could keep it by examining the Persistent" data in localstorage, good idea

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
steinsiv: I'll leave it there for now, it might be different behaviours per browser, but we should remove it if it behaves consistently

Tests/WitsmlExplorer.Api.Tests/Services/CredentialsServiceTests.cs
steinsiv: I have no strong feelings for either, so I will split and we can follow that going forward 👍

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
steinsiv: Agreed to keep as is for now to be re-assessed in later PR WE-614

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
steinsiv: Agreed in meeting 27.10 to keep as is for now to be re-assessed in WE-614

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
steinsiv: I believe we might manage without this authorization preflight
Let re-assess when the WE-614 is done as this PE will reduce a lot of the cookie handling.

Src/WitsmlExplorer.Frontend/services/credentialsService.ts
steinsiv: Implemented as described above

janburak (approved):
- [X] Read issue and PR
- [X] Reviewed the code

PR #1627, opened 2022-10-27
WE-614 CredentialsCache

### LINKED ISSUE #1648: Use `TypedResult` from dotnet7
**Is your feature request related to a problem? Please describe.**

The new `Microsoft.AspNetCore.Http.TypedResults` static class is the “typed” equivalent of the existing `Microsoft.AspNetCore.Http.Results class`. 

See: https://devblogs.microsoft.com/dotnet/asp-net-core-updates-in-dotnet-7-preview-4/#typed-results-for-minimal-apis

After NET7 upgrade, we can use `TypedResults` and preserve the concrete type information.

**Additional context**
See return values in [HttpHandlers](https://github.com/equinor/witsml-explorer/tree/main/Src/WitsmlExplorer.Api/HttpHandlers)

### LINKED ISSUE #1659: Use single session ID with credentials cache
**Is your feature request related to a problem? Please describe.**

Use credential cache in backend mapped to unique ID by either:
 Basic auth - cookie and id
 OAuth - unique id in Bearer token

# Overview
image

steinsiv:
## Fixes
This pull request fixes #1648, #1659 and `WE-614`

## Description
Use a single unique ID for cookies and credentials cache.
- Use `sub` from token when in OAuth mode
- Use `uuid` when in Basic Auth mode 
- Added target and source witsml account used for jobs in `JobInfo`
- Add backend `CredentialCache` with encrypted credentials linked to user id
- Check for valid credentials and witsmlclient in endpoints. Return 401 if not
- Add description on Header in Swagger to explain use
- Use TypedResult over Results, see #1648
-
## Type of change

* This change requires a documentation update

Api, Frontend

*Communication*
* [x] I have made corresponding changes to the documentation

*Code quality*

*Test coverage*
* [x] Existing tests pass

## Further comments


Src/WitsmlExplorer.Api/Extensions/HttpContextExtensions.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+
+using Microsoft.AspNetCore.Http;
+
+using WitsmlExplorer.Api.HttpHandlers;
+namespace WitsmlExplorer.Api.Extensions
+{
+    public static class HttpContextExtensions
+    {
+        public static string GetOrCreateWitsmlExplorerCookie(this HttpContext httpContext, string environment)
```
janburak: Why is the cookie dependent on the environment?
Also would it be better to have IConfiguration as an argument rather than a string, given that the line
> string environment = configuration["ENVIRONMENT"].ToLower(System.Globalization.CultureInfo.CurrentCulture);

has to be called every time?

Src/WitsmlExplorer.Api/HttpHandlers/EssentialHeaders.cs
```diff
@@ -1,50 +1,43 @@
...
-        }
-        public string GetCookie(string server)
+
+        public string GetCookieValue()
         {
-            return (server == WitsmlTargetServer) ? TargetServerCookie : SourceServerCookie;
+            return WitsmlExplorerCookie;
         }
-        public string GetHost(string server)
+        public string GetHeaderValue(string witsmlServer)
```
janburak: Would it be better to make TargetServer and SourceServer public to make retrieving them cleaner than calling GetHeaderValue with the static strings?

Src/WitsmlExplorer.Api/Extensions/HttpContextExtensions.cs
steinsiv: Thanks I will try to check if we can make these changes.
The reason for including environment in the cookie name was to better distinguish and not overwrite the cookie if testing different environments. Having the same name might not be a big issue though

Src/WitsmlExplorer.Api/HttpHandlers/EssentialHeaders.cs
steinsiv: Yes I think so too, will change that

janburak: Also I feel that the name GetHeaderValue does not indicate that it can be only used to retrieve either the TargetServer or the SourceServer header.

Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+using System.Collections.Generic;
+using System.Runtime.Caching;
+
+using Microsoft.Extensions.Logging;
+namespace WitsmlExplorer.Api.Services
+{
+
+    public interface ICredentialsCache
+    {
+        void Set(string cacheId, string encryptedCredentials, CacheItemPolicy policy);
```
janburak: I am getting the following suggestion here:
"Rename virtual/interface member ICredentialsCache.Set(string, string, CacheItemPolicy) so that it no longer conflicts with the reserved language keyword 'Set'. Using a reserved keyword as the name of a virtual/interface member makes it harder for consumers in other languages to override/implement the member. [WitsmlExplorer.Api]csharp(CA1716)".
Should the Set and Get methods be renamed?

Src/WitsmlExplorer.Api/Services/CredentialsService.cs
```diff
@@ -108,39 +92,127 @@ private string Decrypt(string inputString)
...
+                {
+                    CacheCredentials(GetClaimFromTokenValue(token, SUBJECT), result, 1.0);
+                }
             }
-            return new ServerCredentials();
+            return result;
         }
-        private static string GetTokenUserPrincipalName(string token)
+
+        public ServerCredentials GetCredentialsFromCache(bool useOauth, IEssentialHeaders headers, string server, Func<string, string> delDecrypt = null)
```
janburak: delDecrypt seems to never be set, should it be removed from the signature?

Src/WitsmlExplorer.Api/Services/CredentialsService.cs
```diff
@@ -108,39 +92,127 @@ private string Decrypt(string inputString)
...
             }
         }
-        private async Task<ServerCredentials> GetSystemCredentialsWithToken(string token, Uri server)
+
+        public void RemoveCachedCredentials(string clientId)
+        {
+            _credentialsCache.RemoveAllClientCredentials(clientId);
+        }
+
+        public void CacheCredentials(string clientId, ServerCredentials credentials, double ttl, Func<string, string> delEncrypt = null)
```
janburak, line 102: delDecrypt seems to never be set here either, should it be removed from the signature?

janburak: I can now see that it is used by a test to disable encrypting.

Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+    {
+        private readonly ObjectCache _cache = MemoryCache.Default;
+        private readonly ILogger<CredentialsCache> _logger;
+
+        public CredentialsCache(ILogger<CredentialsCache> logger)
+        {
+            _logger = logger;
+        }
+
+        public void Set(string cacheId, string encryptedCredentials, CacheItemPolicy policy)
```
janburak: Would it be better to pass the "keep" boolean here so that the cache can decide the policy based on it?

Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+        public string Get(string cacheId)
+        {
+            return _cache.Get(cacheId) as string;
+        }
+        public long Count()
+        {
+            return _cache.GetCount();
+        }
+
+        public void RefreshSession(string cacheId, string encryptedCredentials)
```
janburak: The RefreshSession method seems to be unused, should it be removed? Also it seems to replicate the [SlidingExpiration policy](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.caching.cacheitempolicy.slidingexpiration?view=dotnet-plat-ext-7.0#system-runtime-caching-cacheitempolicy-slidingexpiration) .

Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
steinsiv: Probably, I did not get the same warning so thanks for the tip

Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
steinsiv: SlidingExpiration was the intended functionality. I see now that it is a perfect fit to use `SlidingExpiration` over `AbsoluteExpiration` so I will change that, thanks a lot :)

Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
steinsiv: Changed to `Set/GetItem`

Src/WitsmlExplorer.Api/Services/CredentialsService.cs
```diff
@@ -108,39 +92,127 @@ private string Decrypt(string inputString)
...
-            ServerCredentials witsmlCredentials = GetCredentialsCookieFirst(headers, EssentialHeaders.WitsmlTargetServer).Result;
-            return (GetTokenUserPrincipalName(headers.Authorization), witsmlCredentials.UserId);
+            ServerCredentials result;
+            if (string.IsNullOrEmpty(headers.GetCookieValue()) || string.IsNullOrEmpty(headers.GetHeaderValue(server)))
+            {
+                result = new ServerCredentials();
+            }
+            else
+            {
+                string host = new Uri(headers.GetHeaderValue(server)).Host;
```
janburak: Doesn't this bit pretty much duplicate GetCredentialsFromCache?

Src/WitsmlExplorer.Api/HttpHandlers/EssentialHeaders.cs
steinsiv: We does not consider any other headers to be `essential` (except for the Authorization ofcourse) so this was the reasoning behind.
In this case using GetHeaderValue ended up being a bit overkill so I will remove that

Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
steinsiv: Not sure, it seems to be a little tight connected to the format of the endpoint query then? 
But we could improve by accepting TTL as a number and then initialize the policy from within the CredentialsCache.
I will implement that for now, what do you think?

Src/WitsmlExplorer.Api/Services/CredentialsService.cs
steinsiv: I think you are correct. They probably started out on different places and have now the same logic. `GetCredentialsFromCache` can be used to fulfill `GetWitsmlCredentialsByCookie`. I will verify and remove if so.

Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
janburak: _Not sure, it seems to be a little tight connected to the format of the endpoint query then?_
Well I looked at it in a way that in practice we use two cache policies, and which policy is used is based on the _keep_ variable.  I think it is barely important anyway given that we construct the policy in only one place (CredentialsService). Any refactoring should rather be decided with regards to making it easy to adjust the policies in code, such as for the SlidingExpiration suggestion.

steinsiv: The `SlidingExpiration` was great. This has been implemented now 👍  I think I will keep the `ttl` approach for now because it seems to be a little more generic and opening up for more than two states for the cache items through the same method.

Src/WitsmlExplorer.Api/Services/CredentialsService.cs
steinsiv: Was able to remove both `GetWitsmlCredentialsByCookie` and `GetWitsmlCredentialsByToken` now so thanks! :)

Src/WitsmlExplorer.Api/HttpHandlers/AuthorizeHandler.cs
```diff
@@ -12,36 +12,38 @@ namespace WitsmlExplorer.Api.HttpHandlers
...
-        public static async Task<IResult> Authorize([FromQuery(Name = "keep")] bool keep, [FromServices] ICredentialsService credentialsService, HttpContext httpContext)
+        public static async Task<IResult> Authorize([FromQuery(Name = "keep")] bool keep, [FromServices] ICredentialsService credentialsService, IConfiguration configuration, HttpContext httpContext)
         {
-            string basicAuth = httpContext?.Request.Headers[EssentialHeaders.WitsmlTargetServer].ToString();
+            bool useOAuth2 = StringHelpers.ToBoolean(configuration[ConfigConstants.OAuth2Enabled]);
+            EssentialHeaders eh = new(httpContext?.Request);
+            string basicAuth = eh.TargetServer;
             ServerCredentials creds = await credentialsService.GetCredentialsFromHeaderValue(basicAuth);
-
+            await credentialsService.VerifyCredentials(creds);
```
janburak, line 21: Should the exception thrown in VerifyCredentials by witsmlClient.TestConnectionAsync() be handled here to give the user a reasonable error message in frontend?

Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
```diff
@@ -0,0 +1,61 @@
...
+            foreach (KeyValuePair<string, object> item in _cache)
+            {
+                if (item.Key.StartsWith(clientId))
+                {
+                    _cache.Remove(item.Key);
+                }
+            }
+        }
+
+        public void LogCache()
```
janburak, line 53: Unused method, any use for it?

steinsiv: Often during development, but I removed any leftover DBG statements in the code before push. Would like to keep it for now

Src/WitsmlExplorer.Api/HttpHandlers/AuthorizeHandler.cs
steinsiv: That might be a good idea yes to catch and give a 401 response. I will check thanks

steinsiv: Seems like it is already being handled by catching an `MessageSecurityException` in `ExceptionMiddleware` giving 401 with an ok message to the end user.

Something similar could perhaps be used for the handlers missing valid creds in cache 🤔

steinsiv: Removed the `HasClient`. Added handling in Middleware, thanks

janburak (approved):
- [X] Read issue and PR
- [X] Reviewed the code

PR #1632, opened 2022-11-01
WE-624 Compare Log to Server

janburak:
## Fixes
This pull request fixes WE-624

## Description
* Implement compare log to server functionality with a modal to show the differences between log curve infos.
* Move "missing object modals" out to their own file to make them reusable.
* Add missing divider to LogCurveInfoContextMenu.
* Extract ProgressSpinner from WellProgress to make the styled (centered) spinner reusable with custom messages.
* Change LogCurveInfo.minDateTimeIndex and LogCurveInfo.maxDateTimeIndex types to string to reflect that they have always been strings and not actually Dates.

Checklist of cases to test while reviewing:
* all mnemonics equal
* target log not found
* log types mismatched (quite of an improbable edge case, might be too cumbersome to test)
* source mnemonic exists, target does not
* target mnemonic exists, source does not
* different start index
* different end index
* depth logs compared correctly
* time logs compared correctly

Well "NO 34/8 V-1 HT2 CompareToServer Test" on Equinor Sitecom Test was used to check correctness of implementation against Production.

## Type of change

* Frontend

## Checklist:

*Communication*

*Code quality*

*Test coverage*
* [x] Existing tests pass


steinsiv (changes requested):
As far as I can tell the functionality works well and as expected. 👍 

I think the GUI for the dialog deviates too much from anything else done previous in witsml-explorer. Especially the header:
image

I would propose to follow the convention used earlier in other dialogs. 

Screenshot 2022-11-15 at 07 21 54

A suggestion is also to split and move the elaborate text into tooltips close to the information they are describing, like below

image

Src/WitsmlExplorer.Frontend/components/Modals/LogComparisonModal.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+  return (
+    <ModalDialog
+      heading={`Log comparison`}
+      content={
+        <>
+          {(indexesToShow && (
+            <>
+              <span>
+                <p>Source well name: {sourceLog.wellName}</p>
```
steinsiv: There should not be any need to specify "`Source` Well", `Wellbore` or `Log`. They are being represented for both source and target

janburak: _Source_ well, wellbore and log names are _expected_ to match the target, but given that a match is based on UID, it is possible that they differ. An alternative I can see is to show both the source and target names, either always or only when they differ (thus having only one field per name).

janburak:
Ref the GUI comment, I think using fields instead of paragraphs looks way better! The same could be used for the "missing object modals".
I am unsure about moving the description to header tooltips. The description concerns the whole table, not individual colums. Also the amount of text would be quite large for a tooltip. I can try and see how it looks.

Src/WitsmlExplorer.Frontend/components/Modals/LogComparisonModal.tsx
steinsiv: Ah I see. Yes then it might make sense to have both displayed or at least include the Source tag

steinsiv:
> Ref the GUI comment, I think using fields instead of paragraphs looks way better! The same could be used for the "missing object modals". I am unsure about moving the description to header tooltips. The description concerns the whole table, not individual colums. Also the amount of text would be quite large for a tooltip. I can try and see how it looks.

I see what you mean yes. I have no obvious solution, maybe the comments regarding `mnemonics` 
```
"Mnemonics that are found in only one of the logs are also included. " 
"Missing mnemonics are indicated by dashes as their index values. " 
"Mnemonics that have a logCurveInfo element but have empty indexes are specifed as undefined. "
"Mnemonics that have equal index values are not shown. " 
```
can be extracted to separate tooltip just for the header `Curve mnemonic`. This leaves one tooltip comment for the headers `source/target/start/end`
```
"Differing index values are shown in bold."
```
And then the rest 
```
The following comparison is based on the logCurveInfo elements and does not look at the logData element itself. "
"The table shows only the mnemonics where the indexes do not match. "
```
could be an ingress to the table? But this is just one suggestion probably not the best one

steinsiv (approved):
Tested manually the following cases 👍 

X all mnemonics equal
X target log not found
log types mismatched (quite of an improbable edge case, might be too cumbersome to test)
X source mnemonic exists, target does not
X target mnemonic exists, source does not
X different start index
X different end index
X depth logs compared correctly
X time logs compared correctly

- [x] Read issue and PR
- [/] Reviewed the code

PR #1673, opened 2022-11-21
WE-671 time zone support in content views

janburak:
## Fixes
This pull request fixes WE-671

## Description
* Add functionality of picking a time zone which will affect how times are shown in content views.
* Change DateTime types in frontend objects to string because they were always strings.
* Change DateTime types in API so that the raw dates from WITSML servers are preserved.

## Type of change

* API, Frontend

## Checklist:

*Communication*

*Code quality*

*Test coverage*
* [x] Existing tests pass

## Further comments
* Display in modals and modification of dateTimes will be addressed in WE-672.
* Use of time zones when opening curves and adjust range will be addressed in WE-681.
* Use of time zones in log comparison will be addressed in WE-682.
* Dates for wells and wellbores were unchanged due to them not showing time.
* date-fns and date-fns-tz packages were used instead of 'moment' due to 'moment' being in maintenance mode. Use of 'moment' would require adding a plugin, thus increasing dependence on 'moment', which should be switched away from.


steinsiv (approved):
- [x] Read issue and PR
- [x] Reviewed the code

PR #1745, opened 2023-02-03
WE-789 Rework cache to store creds in a cacheId-server hierarchy

janburak:
## Fixes
This pull request fixes WE-789

## Description
Store credentials in cache based on the following hierarchy:
cacheId -> Dictionary<serverUrl, Dictionary> -> Dictionary<WITSML username, WITSML password>.
Check whether a header cookie id already exists in cache before using it to store credentials.
Replace the name of clientId with cacheId for consistency.

## Type of change

* API

## Checklist:

*Communication*

*Code quality*

*Test coverage*
* [x] Existing tests pass

### DIFF
--- a/Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
+++ b/Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
@@ -8,11 +8,12 @@ namespace WitsmlExplorer.Api.Services
 
     public interface ICredentialsCache
     {
-        void SetItem(string cacheId, string encryptedPassword, double ttl, string username);
-        public Dictionary<string, string> GetItem(string cacheId);
+        void SetItem(string cacheId, Uri serverUrl, string encryptedPassword, double ttl, string username);
+        public Dictionary<string, Dictionary<string, string>> GetItem(string cacheId);
+        public Dictionary<string, string> GetItem(string cacheId, Uri serverUrl);
         public long Count();
         public void Clear();
-        public void RemoveAllClientCredentials(string clientId);
+        public void RemoveAllClientCredentials(string cacheId);
     }
 
     public class CredentialsCache : ICredentialsCache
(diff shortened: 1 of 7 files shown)


Tests/WitsmlExplorer.Api.Tests/Services/CredentialsCacheTests.cs
```diff
@@ -20,42 +21,97 @@ public CredentialsCacheTests()
...
+            _credentialsCache.RemoveAllClientCredentials(cacheId);
             long after = _credentialsCache.Count();
-            Assert.Equal(3, before);
+            Assert.Equal(2, before);
             Assert.Equal(1, after);
             _credentialsCache.Clear();
         }
+
         [Fact]
         public void Clear_SetThree_ReturnZero()
```
steinsiv: Maybe the signature should be changed a bit because we set two duplicates and one unique cache item resulting in just setting two. and asserting two.
Maybe either named `Clear_SetTwo_ReturnZero` or `Clear_SetThreeWithTeoDuplicates_ReturnZero` ? What do you think?

Tests/WitsmlExplorer.Api.Tests/Services/CredentialsCacheTests.cs
```diff
@@ -20,42 +21,97 @@ public CredentialsCacheTests()
...
+            string cacheId = Guid.NewGuid().ToString();
             string url = "https://somehost";
             string witsmlUsername = "user";
 
             _credentialsCache.Clear();
-            _credentialsCache.SetItem($"{clientId}@{url}{_random.Next(1000)}.com", $"DUMMY_VALUE{_random.Next(1000)}", 1.0, witsmlUsername);
-            _credentialsCache.SetItem($"{clientId}@{url}{_random.Next(1000)}.com", $"DUMMY_VALUE{_random.Next(1000)}", 1.0, witsmlUsername);
-            _credentialsCache.SetItem($"{Guid.NewGuid()}@{url}{_random.Next(1000)}.com", $"DUMMY_VALUE{_random.Next(1000)}", 1.0, witsmlUsername);
+            _credentialsCache.SetItem(cacheId, new Uri($"{url}{_random.Next(1000)}.com"), $"DUMMY_VALUE{_random.Next(1000)}", 1.0, witsmlUsername);
+            _credentialsCache.SetItem(cacheId, new Uri($"{url}{_random.Next(1000)}.com"), $"DUMMY_VALUE{_random.Next(1000)}", 1.0, witsmlUsername);
```
steinsiv: I think we can remove one of these now, because the cache-key does not contain the random number anymore it will just overwrite the other item with the same cacheId.
Also we may change the signature to something like `...OneOfTwo_ReturnOne`

janburak: I can remove one.

Tests/WitsmlExplorer.Api.Tests/Services/CredentialsCacheTests.cs
janburak: I don't think the input matters much when we test the clear method. I can adjust it.

Src/WitsmlExplorer.Api/Services/CredentialsCache.cs
```diff
@@ -8,11 +8,12 @@ namespace WitsmlExplorer.Api.Services
 
     public interface ICredentialsCache
```
steinsiv, line 9: Being a three level dictionary, I think we could consider outlining the cache setup. If you agree we could perhaps add that as a `<summary>` above this class. 
(Something similar to)
```
{ "cacheId" : {
    "serverUrl.Host": {
      "username": "encrypted_password"
    }
  }
}
```

steinsiv (approved):
- [x] Read issue and PR
- [x] Reviewed the code

Issue #1799, opened 2023-03-20
💡 Option to Copy Well/Wellbore when using Copy Log Feature


ian22322232:
As a user it will be convienent to have option to automatically create the parent well and wellbore object if it doesn't exist when using the Copy Log feature. 

Ideally, when a user tries to perform the Copy Log feature there is an option if the well & wellbore doesn't exist to also copy those objects so the Copy Log feature can continue.

PR #1821, opened 2023-04-03
Support copying a log's parent well and wellbore

JoelDavidLang:

## Fixes

#1799 

## Description

This is a draft for supporting copying the parent well and wellbore when an object is copied to another server.

What is implemented:
* Copy well job
* Copy wellbore job
* Copy log with parent job
* Frontend support for copying log with parent
* Update to well elevation model to fix error when copying well

What is not yet implemented:
* Copy objects with parent job
* Frontend support for copying objects with parent
* Unit tests for the new jobs

I would like feedback on whether the direction of the code is the correct one so that I may continue with the not yet implemented changes.

## Type of change

* [ ] Bugfix

* [x] Fronted
* [x] API
* [x] WITSML
* [ ] Other (please describe)

## Checklist:

*Communication*

*Code quality*
* [ ] No new warnings are generated

*Test coverage*

## Further comments

Following the pattern used by the ReplaceLogDataWorker class I implemented the new functionality by adding a new CopyLogWithParent job type that combines the CopyWell, CopyWellbore, and CopyLog jobs into one job.

The frontend was changed so that rather than showing a modal dialog to indicate that a wellbore does not exist on the target server, the dialog asks the user if they want the parent well and wellbore to be copied with the log.

I had to make a change to the Witsml.WellDatum class to properly support copying elevation because I was getting an error about required attributes missing during testing.

Finally, while testing I found that copying logs that are empty or have a decreasing index direction did not work. I would like to implement support for those in a separate PR.


janburak:
Thank you for the PR. 
I took a look at the code and tried it out on our test server. I had to make some changes to make it work. The changes can be found here: https://github.com/janburak/witsml-explorer/commit/20f3536ed2ad4c4ef921d3c39014a32686f03a44. Some things to notice regarding the changes I made:
* The preexisitng WitsmlWell, WitsmlWellbore, and WitsmlLog classes may lack some elements, which would result in data being lost during copying. We might need a separate PR to include all elements from the WITSML specification to ensure that no data is lost. 
* CopyLogJob and CopyObjectsJob are the same so I think it is not necessary to have both of them. The changes I made in the CopyJobs.cs file are a suggestion and there may be a clearer way to set up the jobs.

Considering the following points "_What is not yet implemented: Copy objects with parent job, Frontend support for copying objects with parent_", I think it would be better to implement them in a separate PR to keep this one focused on the CopyLogWithParent job.

In general I think the implemented changes fit well with the repository. I have some suggestions regarding potential minor code improvements but I can wait with those until the PR is out of the draft stage.

I would be glad to answer any questions you might have.

JoelDavidLang:
> I took a look at the code and tried it out on our test server. I had to make some changes to make it work.

Just to be clear, the changes you had to make to get it working are related to the WITSML well and wellbore classes being incomplete in relation to the WITSML 1.4 schema?

When I tested the functionality, it did work for me with what elements are currently in the WITSML classes.

I agree that the classes are incomplete as they are now, and a new PR would be needed to correct this. I assume the current classes were made by hand. Has any thought been given to using code generation based on the XML schema?

> CopyLogJob and CopyObjectsJob are the same so I think it is not necessary to have both of them. The changes I made in the CopyJobs.cs file are a suggestion and there may be a clearer way to set up the jobs.

Do you mean using CopyObjectsJob class in both cases, but using a different JobType to determine which worker class is used?

One thing I notice is that the worker classes use an ILogger<T> where the type is the job class rather than the worker class. I think it would lead to confusing logs if I used the same job class for two different workers.

janburak:
> Just to be clear, the changes you had to make to get it working are related to the WITSML well and wellbore classes being incomplete in relation to the WITSML 1.4 schema?

The complete list of changes that made copying work on my end is:
* Sorted WitsmlWell and WitsmlWellbore variables according to the specification. The Sitecom server I tested against did not accept queries with arbitrary ordering of elements.
* Corrected the type of WitsmlWell.WaterDepth from string to WitsmlLengthMeasure.
* Replaced the use of CopyLogJob in CopyLogWithParentJob with using the Source and Target of ICopyJob (I was getting a null error because Source and Target were not set). An alternative here would be to make CopyLogWithParentJob implement Job directly, which would make sense if one would want to adjust the Description of the CopyLogWithParentJob.

> Has any thought been given to using code generation based on the XML schema?

I have not tried it. I use a set of regex queries to convert XSD to C# classes and then adjust the types based on conventions in the Witsml project.

> Do you mean using CopyObjectsJob class in both cases, but using a different JobType to determine which worker class is used?
One thing I notice is that the worker classes use an ILogger where the type is the job class rather than the worker class. I think it would lead to confusing logs if I used the same job class for two different workers.

Good point. It's a bit unfortunate that the ILogger category comes from the job class. In that case it does make sense to have a 1 to 1 mapping between Jobs and Workers. The JobType does determine which worker is used when calling the `api/jobs/{jobType}` route. The Job classes define the required input to the various Workers, as well as mediate monitoring of jobs using JobInfo.

Will you be joining the Equinor-Petrolink meeting on tuesday 11.04 with Ian Lee? I think it would be good for both of us to attend it to establish communication channels.

JoelDavidLang:
I've incorporated the changes you suggested and marked this PR ready for review.

I do have a question about Prettier though. The Prettier check on the frontend would not pass unless I removed the indentation I had done for the HTML element in the CopyWellboreModal.tsx file. But I noticed that other modal dialog files had this indentation and were not flagged by Prettier when the check is done on GitHub, so I'm confused about why that was necessary.

> Will you be joining the Equinor-Petrolink meeting on tuesday 11.04 with Ian Lee? I think it would be good for both of us to attend it to establish communication channels.

Yes, I'll be there.

janburak:
I just let the [pre-commit hook](https://github.com/equinor/witsml-explorer/blob/main/CONTRIBUTING.md#code-style-guidelines) deal with indentation for me automatically. I think identation is enforced when max line length is exceeded (180 in .editorconfig but I might be wrong). I agree that it is somewhat unnecessary, but wouldn't consider it an important issue.

Src/WitsmlExplorer.Frontend/models/jobs/wellReference.ts
```diff
@@ -0,0 +1,4 @@
+export default interface WellboreReference {
```
janburak: I assume it was supposed to be WellReference instead of WellboreReference?

Src/Witsml/Data/WitsmlWell.cs
```diff
@@ -1,16 +1,30 @@
...
     {
+        [XmlAttribute("uid")]
+        public string Uid { get; set; }
         [XmlElement("name")]
         public string Name { get; set; }
         [XmlElement("code")]
         public string Code { get; set; }
         [XmlElement("elevation")]
-        public string Elevation { get; set; }
+        public WellElevationCoord Elevation { get; set; }
```
janburak: I think WitsmlMeasureWithDatum can be used here, given that it has the same fields as WellElevationCoord.

Src/WitsmlExplorer.Frontend/models/jobs/wellReference.ts
JoelDavidLang: Correct.

Src/Witsml/Data/WitsmlWell.cs
JoelDavidLang: Yes, you're right.

Looking at the schema, I see measuredDepthCoord, wellVerticalDepthCoord, and wellElevationCoord have the same fields, but the difference between them is that the allowed values for uom are slightly different. Since it doesn't seem like there is any need to restrict that uom to an enumeration in code too, then yeah WitsmlMeasureWithDatum makes sense.

That said, looking at the code, I see there is a WitsmlMeasuredDepthCoord class that inherits from WitsmlMeasureWithDatum, but doesn't add any new members. This class is used many places in the code. Should a similar subclassing be done here for WellElevationCoord, or just use WitsmlMeasureWithDatum directly?

Src/WitsmlExplorer.Frontend/components/Modals/CopyWellboreModal.tsx
```diff
@@ -0,0 +1,10 @@
+import OperationType from "../../contexts/operationType";
+import { DispatchOperation } from "../ContextMenus/ContextMenuUtils";
+import ConfirmModal from "./ConfirmModal";
+
+export function displayCopyWellboreModal(wellboreUid: string, dispatchOperation: DispatchOperation, onConfirm: () => void) {
+  const content = <span>The wellbore {wellboreUid} does not exist on the target server. Do you want to copy it and its parent well?</span>;
+
+  const confirmation = <ConfirmModal heading={`Copy wellbore?`} content={content} onConfirm={onConfirm} confirmColor={"danger"} confirmText={`Copy`} switchButtonPlaces={true} />;
```
janburak: `confirmColor={"danger"}` is usually used to warn the user that something will be deleted. `confirmColor={"primary"}` might be a better choice here.

Src/Witsml/Data/WitsmlWell.cs
janburak: I do not think there is any benefit to using Measure classes that do not add new members, given that we do not have any plans to handle the different kinds of measures based on the derived types. I assume that WitsmlMeasureWithDatum should be sufficient for now. As long as the required fields are present in the class it should be fine either way.

Src/WitsmlExplorer.Api/Workers/Copy/CopyLogWithParentWorker.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                (WorkerResult result, RefreshAction) wellboreResult = await _copyWellboreWorker.Execute(job.CopyWellboreJob);
+
+                if (!wellboreResult.result.IsSuccess)
+                {
+                    return wellboreResult;
+                }
+            }
+
+            return await _copyLogWorker.Execute(new() { Source = job.Source, Target = job.Target });
```
janburak: `_copyLogWorker` returns a refreshAction that refreshes the logs on a given wellbore. This is insufficient when the worker copies over the well or the wellbore. I think the following would ensure correct refreshing:
```
       public override async Task<(WorkerResult WorkerResult, RefreshAction RefreshAction)> Execute(CopyLogWithParentJob job)
        {
            RefreshAction refreshAction = null;
            if (job.CopyWellJob != null)
            {
                (WorkerResult result, RefreshAction refreshWell) wellResult = await _copyWellWorker.Execute(job.CopyWellJob);
                refreshAction = wellResult.refreshWell;
                if (!wellResult.result.IsSuccess)
                {
                    return wellResult;
                }
            }
            if (job.CopyWellboreJob != null)
            {
                (WorkerResult result, RefreshAction refreshWellbore) wellboreResult = await _copyWellboreWorker.Execute(job.CopyWellboreJob);
                refreshAction ??= wellboreResult.refreshWellbore;
                if (!wellboreResult.result.IsSuccess)
                {
                    return wellboreResult;
                }
            }
            (WorkerResult result, RefreshAction refreshLogs) = await _copyLogWorker.Execute(new() { Source = job.Source, Target = job.Target });
            refreshAction ??= refreshLogs;
            return (result, refreshAction);
        }
```
       
BTW there is currently a bug in frontend regarding refreshing, that can be fixed by adding `wells` to the dependency list in RefreshHandler.tsx on line 99.

JoelDavidLang:
Recommended changes have been pushed.

janburak (approved):
- [X] Read issue and PR
- [X] Reviewed the code

Give this comment a 👍 if the PR is ready to be merged in.

PR #1822, opened 2023-04-07
WE-843 optimize LogDataReader

janburak:

## Fixes

This pull request fixes WE-843

## Description

* Optimize LogDataReader by buffering query results while running on a separate thread.

## Type of change

* [ ] Bugfix

* [ ] Fronted
* [x] API
* [ ] WITSML
* [ ] Other (please describe)

## Checklist:

*Communication*

*Code quality*

*Test coverage*

## Further comments


Src/WitsmlExplorer.Api/Services/LogDataReader.cs
```diff
@@ -103,5 +121,35 @@ public async Task<WitsmlLogData> GetNextBatch()
...
+            if (_finished && _buffer.Count == 0)
+            {
+                return null;
+            }
+            return await WithTimeout(_buffer.ReceiveAsync());
+        }
+
+        private async Task<T> WithTimeout<T>(Task<T> task)
+        {
+            int timeout = 5 * 60 * 1000;
```
HakonHesselberg: This can be a const

```suggestion
            const int timeout = 5 * 60 * 1000;
```

Src/WitsmlExplorer.Api/Services/LogDataReader.cs
```diff
@@ -28,24 +29,29 @@ public class LogDataReader
...
         /// </summary>
         /// <param name="witsmlClient">The witsmlClient used to fetch data</param>
         /// <param name="sourceLog">A WitsmlLog object used to retrieve well, wellbore, and log uids, and the start, end, type, and mnemonic of the index curve.</param>
         /// <param name="mnemonics">A list of mnemonics to fetch. The index curve will be added to this list if not present.</param>
         /// <param name="logger">A logger or null</param>
-        public LogDataReader(IWitsmlClient witsmlClient, WitsmlLog sourceLog, List<string> mnemonics, ILogger logger)
+        /// <param name="bufferSize">How many query results to buffer at a time. Defaults to 4.</param>
+        public LogDataReader(IWitsmlClient witsmlClient, WitsmlLog sourceLog, List<string> mnemonics, ILogger logger, int bufferSize = 4)
         {
             _witsmlClient = witsmlClient;
```
HakonHesselberg: Consider adding a null-check here:
_witsmlClient = witsmlClient ?? throw new ArgumentNullException(nameof(witsmlClient));
_uidWell = sourceLog?.UidWell ?? throw new ArgumentNullException(nameof(sourceLog));

Src/WitsmlExplorer.Api/Services/LogDataReader.cs
```diff
@@ -63,23 +69,36 @@ public LogDataReader(IWitsmlClient witsmlClient, WitsmlLog sourceLog, List<strin
...
             }
+            Produce();
         }
 
-        /// <summary>
-        /// Fetches and returns the next batch of log data.
-        /// </summary>
-        /// <returns>WitsmlLogData or, if there is no more data to fetch, null</returns>
-        public async Task<WitsmlLogData> GetNextBatch()
+        private async void Produce()
```
HakonHesselberg: [Best practice](https://learn.microsoft.com/en-us/archive/msdn-magazine/2013/march/async-await-best-practices-in-asynchronous-programming) is to prefer Task as return type instead of void. Is there a reason to use void here?

(Note that the call to `Produce();` above needs to be wrapped in a `Task.Run(Produce)` if the return type is changed.

Src/WitsmlExplorer.Api/Services/LogDataReader.cs
```diff
@@ -103,5 +121,35 @@ public async Task<WitsmlLogData> GetNextBatch()
...
+                throw _exception;
+            }
+            if (_finished && _buffer.Count == 0)
+            {
+                return null;
+            }
+            return await WithTimeout(_buffer.ReceiveAsync());
+        }
+
+        private async Task<T> WithTimeout<T>(Task<T> task)
```
HakonHesselberg: I think this method can be replaced by using a `cancellationToken` with a timeout.

Create a field:
` private readonly CancellationTokenSource _cancellationTokenSource;`

In the constructor:
```
const int timeout = 5 * 60 * 1000;
_cancellationTokenSource = new CancellationTokenSource(timeout);
```

In the Produce method:
`await _buffer.SendAsync(logData, _cancellationTokenSource.Token);`

In the GetNextBatch method:
`return await _buffer.ReceiveAsync(_cancellationTokenSource.Token);`

This should handle the timeout in a similar way, while also providing a cancellation token if there are other needs.

HakonHesselberg (changes requested):
Looks good overall, added a few suggestions.

Src/WitsmlExplorer.Api/Services/LogDataReader.cs
janburak: Added

Src/WitsmlExplorer.Api/Services/LogDataReader.cs
janburak: Timeout was removed

Src/WitsmlExplorer.Api/Services/LogDataReader.cs
janburak: No reason to use void. Replaced with Task.Run

Src/WitsmlExplorer.Api/Services/LogDataReader.cs
janburak: Rewritten to use cancellationTokenSource and IAsyncDisposable instead of timeout.

HakonHesselberg (approved):

Issue #1209, opened 2022-05-20
🐛 `/api/wells/<id>` error in `WellDatums` and `WellLocations`


steinsiv:
**:bug: Describe the bug**
See discussion in https://github.com/equinor/witsml-explorer/pull/791#issuecomment-953541763

Example Well: `/api/wells/MzBfOS1wLTNodDI=`

From Witsml we get (excerpt):

```xml
    <wellDatum uid="DF">
      <name>Derrick Floor</name>
      <code>DF</code>
      <elevation uom="m" datum="SL">29</elevation>
    </wellDatum>
    <wellDatum uid="SL">
      <name>Sea Level</name>
      <code>SL</code>
    </wellDatum>

    <groundElevation uom="m">0</groundElevation>
    <waterDepth uom="m">104</waterDepth>
    
    <wellLocation uid="0">
      <latitude uom="dega">60.46</latitude>
      <longitude uom="dega">2.7</longitude>
    </wellLocation>
    <wellLocation uid="1">
      <easting uom="m">48.58</easting>
      <northing uom="m">67.27</northing>
    </wellLocation>
    <wellLocation uid="2">
      <localX uom="m">48.58</localX>
      <localY uom="m">67.27</localY>
    </wellLocation>
```

but returned from the API we only get this:

```json
  "wellDatum": {
    "name": "Derrick Floor",
    "code": "DF",
    "elevation": "29"
  },
  "waterDepth": "104",
  "wellLocation": {
    "latitude": {
      "value": "60.46",
      "uom": "dega"
    },
    "longitude": {
      "value": "2.7",
      "uom": "dega"
    }
  }
```

janburak:
WellLocation in WitsmlWell should be List<WitsmlLocation> instead of WitsmlLocation.

wellLocation, wellDatum and waterDepth are not used in frontend so they should not be sent anyway.

This should probably be fixed as part of implementing the whole WitsmlWell and WitsmlWellbore models ref #1821

PR #2032, opened 2023-08-31
Error in well datums#1209

petrlicman:

## Fixes

This pull request fixes 

## Description

## Type of change

* [x] Bugfix

* [ ] Frontend
* [x] API
* [ ] WITSML
* [ ] Other (please describe)

## Checklist:

*Communication*

*Code quality*

*Test coverage*

## Further comments


Src/WitsmlExplorer.Api/Models/Well.cs
```diff
@@ -13,7 +13,7 @@ public class Well
         public string DateTimeCreation { get; init; }
         public string DateTimeLastChange { get; init; }
         public string ItemState { get; init; }
-        public IEnumerable<Wellbore> Wellbores { get; set; }
+        public IList<Wellbore> Wellbores { get; set; }
```
janburak, line 16: What is the advantage of using IList instead of IEnumerable?

Src/WitsmlExplorer.Api/Services/WellService.cs
```diff
@@ -65,21 +68,21 @@ private async Task<IEnumerable<Well>> GetWellsInformation(string wellUid = null)
                     ItemState = well.CommonData.ItemState,
                     StatusWell = well.StatusWell,
                     PurposeWell = well.PurposeWell,
-                    WellDatum = WellDatum.FromWitsmlWellDatum(well.WellDatum),
+                    WellDatum = WellDatum.FromWitsmlWellDatum(well.WellDatum).FirstOrDefault(),
```
janburak: It makes little sense to return only the first WellDatum or WellLocation when the underlying WITSML model contains a list. The API Well model should either contain lists of WellDatums and WellLocations or not contain them at all. I think that we should remove them from the API model given that we have no use for them in the frontend or any of the workers.

janburak: Also every time one enters the production server, thousands of wells are fetched, so it would be good to keep the return values of GetWellsInformation as small as possible.

Src/WitsmlExplorer.Api/Models/Well.cs
petrlicman: avoiding reenumerate linq queries - as it is inproperly used on many places in this app, where linq chain is reenumerated many times

Src/WitsmlExplorer.Api/Services/WellService.cs
```diff
@@ -65,21 +68,21 @@ private async Task<IEnumerable<Well>> GetWellsInformation(string wellUid = null)
...
                     WaterDepth = well.WaterDepth?.Value,
-                    WellLocation = WellLocation.FromWitsmlLocation(well.WellLocation),
+                    WellLocation = WellLocation.FromWitsmlLocation(well.WellLocation).FirstOrDefault(),
                     Country = well.Country
                 }
                 ).ToList();
-            double elapsed = DateTime.Now.Subtract(start).Milliseconds / 1000.0;
```
janburak, line 74: Good catch :+1:

Src/WitsmlExplorer.Api/Services/WellService.cs
petrlicman: The amount of data where not changed, it was processed in the same way, just mapping was done only for first item. Now all received data in list are mapped, but because of current DTO sent to frontend contains one item only, the first item is used - as it was so, but it was done in the static method already, just not so obvioous

janburak: My point as, as mentioned in the comment of the issue #1209, is that there is no reason to include WellDatum, WellLocation, and WaterDepth in the DTO to begin with, because they are not used in the frontend, and there are no current plans to use them.

Src/WitsmlExplorer.Api/Services/WellService.cs
```diff
@@ -31,24 +33,25 @@ public WellService(IWitsmlClientProvider witsmlClientProvider, IWellboreService
             _wellboreService = wellboreService;
         }
 
-        public async Task<IEnumerable<Well>> GetWells()
+        public async Task<IList> GetWells()
```
eliasbruvik: Looks like you accidentally lost `<Well>`

eliasbruvik (changes requested):

Src/WitsmlExplorer.Api/Services/WellService.cs
petrlicman: yes, you are right, thanks. fixed.

eliasbruvik (approved):

PR #2193, opened 2024-01-08
FIX-2149 Implement log curve priority

### LINKED ISSUE #2149: Ability to set an official RTOF prioritized subset of curves and to filter on this subset💡
RTOF: Real Time Order Form

Today we are getting more and more channels in the WITSML Logs and it is often taking long time to select out the prioritized channels, especially when you want to very quickly look at all the prioritized channels for instance with Streaming turned on. The users define a prioritized subset through an RTOF, and we want to be able to manually set this in WEx.

Ability to manually set  the prioritized subset of channels.  This subset shall be stored in the WEx Cosmos database.

The prioritized subset shall be defined pr. wellbore, so that the well/wellbore identifisers can be used for ID.
The prioritized set shall also work across servers, and across logs.
Example: If a curve GS_BPOS is part of the subset, then it shall be shown in WEx regardless which server or log it is located in.

A user will be able to set prioritized channels by selecting a subset of the curves from a log, right click and select "set priority". Options for edit priority shall also be available when right clicking, and this shall bring up a popup window where you can see the selected curves and also remove curves, and add curves manually (typically curve names which are not yet available). In this window you can also clear the list, and also give the priority set  a name (The name will typically be the section name : example : 26inch order, 17.25inch order etc....). The name shall be visible either as a mouseover when the toggle view is activated, but if space allows it, visible to the right of the slider in the header (in this case a mouseover is not needed).

- If a prioritized set exists already, then new channels will be added to the prioritized list.
- If a user makes a selection of curves he should also have the option to toggle between "Prioritized vs All" (in the header, see below for more details)

When this filter/slider is toggled, all "non prioritized" curves will be hidden.

UX: The user can toggle on / off this filter by using a sliderin the header, the sliderwill by default be turned off, and the label is: "Show only Priority Curves"

This toggle checkbox will not be persistent. The user will always need to toggle it on manually whenever listing curves in a log, and it should be clearly visible to users that the filter is activated to prevent the user from believing that he is viewing all data.

The toggle shall be available above the content view, and it shall be visible and active only when listing the curves available in a log.  

The total number of channels listed in the header should not be changed by making the filter.
Example: If there are 143 curves in a log, and 12 of these are set as a prioritized filter, the total number of curves displayed in the header should still be 143, so inform the user that he has made a filtering choice: 
image

The prioritized subset shall be applicable across servers. 
This means that if you set a subset on one server, and a curve in the subset does is not available on another server, WEx will still look for it, and list it if it becomes available later.

The curve RTOF subset shall be stored in the database so that it will become applicable across user sign-ins.

The user should also have an option to create a "personal" filter which will be stored in the local storage, and also it should be possible to hide all "null mnemonics" which does not contain any data but these will be specified in separate User Stories

eliasbruvik:

## Fixes

This pull request fixes #2149 

## Description

* Added a repository for log curve priority (cosmos or mongo - similar to the server list).
  * The repository uses the id (`wellUid-wellboreUid`) to fetch a mnemonic list.
  * The list of prioritized curves are not unique for each user or server, only well and wellbore.
  * Unlike for the server list, the Admin role is not required to edit prioritized curves.
* Made sure the existing repositories (mongo and cosmos) returns null if trying to fetch an item that does not exist.
* Added routes, a handler and a service to get/create/update/delete priorities.
* ContentTable now uses the id property (if available) of each data element as row id instead of the index. This is to make sure that when removing items from the table, the checked state will be removed instead of keeping the new rows with the same index checked.
* The LogCurveInfoContextMenu (context menu that pops up when right clicking within the edit modal) does not use the ContextMenu component. This is because using the standard way to present a context menu would result in the menu being hidden behind the modal. 
* Removed some style from the content of modals (in `ModalDialog.tsx`). This is because the correct style is already set globally in `GlobalStyles.tsx`, and the removed style did not take into account light or dark mode so the scroll bar always remained white.

## Type of change

* [ ] Bugfix

* [x] Frontend
* [x] API
* [ ] WITSML
* [ ] Other (please describe)

## Checklist:

*Communication*

*Code quality*

*Test coverage*

## Further comments


Src/WitsmlExplorer.Frontend/components/ContextMenus/LogCurveInfoContextMenu.tsx
```diff
@@ -208,15 +243,24 @@ const LogCurveInfoContextMenu = (
...
-          key={"analyzeGaps"}
-          onClick={
-            onClickAnalyzeGaps
-          } /*disabled={checkedLogCurveInfoRows.length !== 1}*/
-        >
+        <MenuItem key={"analyzeGaps"} onClick={onClickAnalyzeGaps}>
           <StyledIcon name="beat" color={colors.interactive.primaryResting} />
           <Typography color={"primary"}>Analyze gaps</Typography>
         </MenuItem>,
+        <MenuItem key={"setPriority"} onClick={onClickSetPriority}>
```
janmarius: It would be nice to make this name remove priority if checked objects are already prioritized. But if one or more checked objects are not yet prioritized, then name it set priority?

Maybe also change the icon to favorite_filled when already prioritized?

Src/WitsmlExplorer.Frontend/components/ContentViews/LogCurveInfoListView.tsx
```diff
@@ -231,6 +260,16 @@ export const LogCurveInfoListView = (): React.ReactElement => {
...
+    </CommonPanelContainer>,
+    <CommonPanelContainer key="showPriority">
+      <Switch
+        checked={showOnlyPrioritizedCurves}
+        disabled={prioritizedCurves.length === 0}
+        onChange={() =>
+          setShowOnlyPrioritizedCurves(!showOnlyPrioritizedCurves)
+        }
+      />
+      <Typography>Show only Priority Curves</Typography>
```
janmarius: Show Only Prioritized Curves?

Src/WitsmlExplorer.Frontend/components/ContextMenus/LogCurvePriorityContextMenu.tsx
```diff
@@ -0,0 +1,55 @@
...
+          ? {
+              top: position.mouseY,
+              left: position.mouseX
+            }
+          : undefined
+      }
+      onContextMenu={preventContextMenuPropagation}
+      colors={colors}
+    >
+      <MenuItem key={"delete"} onClick={onClickDelete}>
```
janmarius, line 46: It was difficult to find how to delete prioritized curves. Could we add a button in the `LogCurvePriorityModal.tsx` modal that gets enabled when objects are checked instead?

janmarius (changes requested):

Src/WitsmlExplorer.Frontend/components/ContextMenus/LogCurvePriorityContextMenu.tsx
eliasbruvik: This is the same way you delete rows from all the other tables, so I don't think it will be a problem. Like here in the curve values view:

image

Src/WitsmlExplorer.Frontend/components/ContentViews/LogCurveInfoListView.tsx
eliasbruvik: I just copied the text specified in the issue, but agree that this sounds better.

Src/WitsmlExplorer.Frontend/components/ContextMenus/LogCurveInfoContextMenu.tsx
eliasbruvik: Good idea!

Src/WitsmlExplorer.Frontend/components/ContextMenus/LogCurvePriorityContextMenu.tsx
janmarius: I know, I just found it a bit hard to find. It is fine for now :)

janmarius (dismissed):
*Reviewer checklist*
- [x] Read issue and PR
- [x] Reviewed the code

janmarius (approved):
*Reviewer checklist*
- [x] Read issue and PR
- [x] Reviewed the code

Issue #2191, opened 2024-01-05
Problem with Copy Curve between servers if you include INDEX curve💡


ThomasHalland:
In this case WEx tries to delete/insert the index curve which results in an error. 
This need to be fixed.

eliasbruvik:
My first thought was that we could simply skip copying the index curve as the target server already should have an index curve. However, even though the index curve exists, and there is no data to copy for the index curve specifically, we still need to copy other properties of the index curve like unit, curveDescription etc. 

#2205 also addresses copying the other properties, but in another setting.

eliasbruvik:
Fixed by #2221

PR #2221, opened 2024-01-26
FIX-2205 Fix copy log curve bugs

### LINKED ISSUE #2205: Copy log curve does not copy LogCurveInfo properties to existing LogCurveInfo objects🐛
When using Copy or Copy WIth Range on a log curve, and pasting it to another log, other properties (like unit and curveDescription) of the LogCurveInfo are not replaced. 

This bug is only related to the copy methods that does not replace the entire LogCurveInfo object.

Steps to reproduce:
1. Navigate to a log you have on 2 servers.
2. Find a curve that exists on both servers.
3. Go into properties of the curve on server A, and change the unit.
4. Click "Copy" on the curve on Server A, and then paste it to the same log on server B.
5. Notice how the unit has not been changed.
6. Delete the curve on server B, and paste it to the log again.
7. Notice how the unit has now been changed.

Expected result:
The properties should be updated after copying even though the curve already exists.

eliasbruvik:

## Fixes

This pull request fixes #2205 and #2191

## Description

Fixed these bugs:
* Error when using "CopyToServer" if you include the indexCurve.
* LogCurveInfo properties are not copied if the target curve already exists.
* LogCurveInfo properties are not copied if using "Copy with range to server", and the curve is outside the selected range.

New behavior:
* If using "CopyToServer" with the indexCurve, it won't attempt to delete it.
* The LogCurveInfo object is always copied for selected curves, even if the target LogCurveInfo object already exists, and even if it's not within the specified range.
* When copying the index curve, properties other than uid and mnemonic in the LogCurveinfo object will be copied, even if the names of the index curves does not match. The uid and mnemonic of index curves can not be changed for a log.

Note for the reviewer:
The removed Assert in the test was removed as for this test there are no "existing" mnemonics to copy. Before I ensured that the target log had the correct mnemonic and uid of the indexCurve LogCurveInfo object, the worker added it as an existing mnemonic. As we do not have any existing mnemonics, there is one less call to UpdateInStore.

## Type of change

* [x] Bugfix

* [x] Frontend
* [x] API
* [ ] WITSML
* [ ] Other (please describe)

## Checklist:

*Communication*

*Code quality*

*Test coverage*

## Further comments

### DIFF
--- a/Src/WitsmlExplorer.Api/Workers/Copy/CopyLogDataWorker.cs
+++ b/Src/WitsmlExplorer.Api/Workers/Copy/CopyLogDataWorker.cs
@@ -2,6 +2,7 @@ using System;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Linq;
+using System.Text.Json;
 using System.Threading.Tasks;
 
 using Microsoft.Extensions.Logging;
@@ -204,7 +205,7 @@ namespace WitsmlExplorer.Api.Workers.Copy
                 WitsmlLogs sourceData = await RequestUtils.WithRetry(async () => await GetSourceWitsmlClientOrThrow().GetFromStoreAsync(query, new OptionsIn(ReturnElements.DataOnly)), Logger);
                 WitsmlLog sourceLogWithData = sourceData?.Logs?.FirstOrDefault();
 
-                if (sourceLogWithData == null)
+                if (sourceLogWithData?.LogData == null)
                 {
                     break;
                 }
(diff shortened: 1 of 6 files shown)


Src/WitsmlExplorer.Api/Workers/Copy/CopyLogDataWorker.cs
```diff
@@ -275,33 +276,50 @@ private static WitsmlData CollateData(List<string[]> oldRows, double index)
...
+            foreach (string mnemonic in sourceMnemonics)
             {
-                if (targetLog.LogCurveInfo.All(lci => !string.Equals(lci.Mnemonic, mnemonic, StringComparison.OrdinalIgnoreCase)))
+                bool isIndexCurve = string.Equals(mnemonic, sourceLog.IndexCurve.Value, StringComparison.OrdinalIgnoreCase);
+                string targetIndexCurveMnemonic = targetLog.IndexCurve.Value;
+
+                var sourceLogCurveInfo = sourceLog.LogCurveInfo.Find(lci => string.Equals(lci.Mnemonic, mnemonic, StringComparison.OrdinalIgnoreCase));
+                var targetLogCurveInfo = targetLog.LogCurveInfo.Find(lci => string.Equals(lci.Mnemonic, isIndexCurve ? targetIndexCurveMnemonic : mnemonic, StringComparison.OrdinalIgnoreCase));
+
+                if (targetLogCurveInfo == null || !LogCurveInfosAreEqual(sourceLogCurveInfo, targetLogCurveInfo))
```
robertbasti, line 288: if (targetLogCurveInfo == null  
and later you are trying:
 sourceLogCurveInfo.Uid = targetLogCurveInfo.Uid;
 sourceLogCurveInfo.Mnemonic = targetLogCurveInfo.Mnemonic;
In case targetLogCurveInfo is null, you will get Null reference exception

Src/WitsmlExplorer.Api/Workers/Copy/CopyLogDataWorker.cs
```diff
@@ -275,33 +276,50 @@ private static WitsmlData CollateData(List<string[]> oldRows, double index)
...
         {
-            List<WitsmlLogCurveInfo> newLogCurveInfos = new();
-            foreach (string mnemonic in sourceMnemonics.Where(mnemonic =>
-            !string.Equals(targetLog.IndexCurve.Value, mnemonic, StringComparison.OrdinalIgnoreCase) &&
-            !string.Equals(sourceLog.IndexCurve.Value, mnemonic, StringComparison.OrdinalIgnoreCase)))
+            List<WitsmlLogCurveInfo> logCurveInfosToUpdate = new();
+            foreach (string mnemonic in sourceMnemonics)
             {
-                if (targetLog.LogCurveInfo.All(lci => !string.Equals(lci.Mnemonic, mnemonic, StringComparison.OrdinalIgnoreCase)))
+                bool isIndexCurve = string.Equals(mnemonic, sourceLog.IndexCurve.Value, StringComparison.OrdinalIgnoreCase);
```
robertbasti, line 282: sourceLog.IndexCurve.Value
targetLog.IndexCurve.Value
Is that assured that IndexCurve is not null in both cases?

Src/WitsmlExplorer.Api/Workers/Copy/CopyLogDataWorker.cs
eliasbruvik: In the case where `isIndexCurve` is true, which is when `targetLogCurveInfo.Uid` is used, `targetLogCurveInfo` will not be null. Logs cannot exist without a index curve, and it should find the index curve even if the name differs between source and target.

If you have any suggestion to make it more clear to the reader that it is safe, let me know.

Src/WitsmlExplorer.Api/Workers/Copy/CopyLogDataWorker.cs
eliasbruvik: `indexCurve` is a required field, so it should not be null.
This is from the xsd:
`<xsd:element name="indexCurve" type="witsml:str32" minOccurs="1" maxOccurs="1">`

robertbasti (approved):

Issue #2195, opened 2024-01-09
Update search view after batch updating


eliasbruvik:
After batch updating an object from the search view, I want the view to refresh the search. 

This only applies when updating a property that shows in the search view, like `name`.

robertbasti:
could you please describe it more details and add some screenshots?

eliasbruvik:
If you select to search for Logs or Rigs (or something else that opens the ObjectSearchListView.
![image](https://github.com/equinor/witsml-explorer/assets/70512270/a73defd5-2bbf-43df-a40f-509817f2b8be)

Then right click on a couple, and select Batch Update.
![image](https://github.com/equinor/witsml-explorer/assets/70512270/b10b578d-5f51-4cc2-b9ca-145a0d758ae4)

Edit the name, and click save. After the batch update is complete, the items (names) in the search view should be updated.
![image](https://github.com/equinor/witsml-explorer/assets/70512270/4831230b-c87b-43ba-9f07-160ad651e412)

Currently in our application, when we start a job in the API that we want to update something in the frontend, we return a RefreshAction. However, the `RefreshHandler.tsx` is not configured to update the search. The data from the search is placed in the `searchResults` variable in the `Filter` object, which is used in ObjectSearchListView through FilterContext.
![image](https://github.com/equinor/witsml-explorer/assets/70512270/2a6f98e2-8010-4eb8-8b78-437472fb042b)

The solution must find a way to let the RefreshHandler know that it's the search view that should be updated, and then update the searchResults by updating the value in the FilterContext accordingly.

Note that the ObjectSearchListView does not fetch all properties for the objects, so we only need to refresh if any of the shown properties are updated. If let's say runNumber is changed in the batchUpdate, but not the name, refreshing the search is unnecessary.

eliasbruvik:
Fixed by #2322

PR #2322, opened 2024-03-19
2195 refresh after batch update

vaclavbasniar:

## Fixes 

[//]: #2195 (Write the GitHub issue number starting with #, or, for Equinor only, the Jira issue number starting with WE-)

This pull request fixes updating of search view after batch update in search view  #2195.

## Description

On front end side:
* Added originJobType to RefreshAction to be able to determine job type in RefreshHandler. 
* Added ObjectOnWellbore list to RefreshAction to get a list of batch updated objects in RefreshHandler. 
* Added #refreshSearchFilter function to RefreshHandler, which invalidates search query and changed objects queries.
On API side:
* Added BatchModifyObjectsOnSearchWorker, which reuses BatchModifyObjectsOnWellboreWorker code with different RefreshAction type - returns list of changed objects.

## Type of change

* [x] Bugfix

* [x] Frontend
* [x] API
* [ ] WITSML
* [ ] Other (please describe)

## Checklist:

*Communication*

*Code quality*

*Test coverage*

## Further comments


Src/WitsmlExplorer.Api/Jobs/BatchModifyObjectsOnSearchJob.cs
eliasbruvik, line 1: I don't think we need this new job and worker. I think it's fine to change the ReturnAction returned from BatchModifyObjectsOnWellboreWorker to your new BatchRefreshObjects instead.

Src/WitsmlExplorer.Api/Models/RefreshAction.cs
```diff
@@ -14,23 +15,26 @@ public enum RefreshType
 
     public abstract class RefreshAction
     {
-        protected RefreshAction(Uri serverUrl, RefreshType refreshType)
+        protected RefreshAction(Uri serverUrl, RefreshType refreshType, JobType originJobType)
         {
             ServerUrl = serverUrl;
             RefreshType = refreshType;
+            OriginJobType = originJobType;
```
eliasbruvik: I think it would be better to not add originJobType. In the refreshHandler, I think we can rather invalidate the search queries for any filterType that has received an update regardless on the job type.

If I for example perform a search, then navigate to the object and change it manually in properties (not batch update), and then go back to the search, it would be nice if it was invalidated so that I get the updated search result.

Src/WitsmlExplorer.Frontend/components/RefreshHandler.tsx
```diff
@@ -40,10 +43,15 @@ const RefreshHandler = (): React.ReactElement => {
...
                   );
                 }
-                if (refreshAction.objectUid == null) {
-                  refreshWellboreObjects(refreshAction);
-                } else {
+                if (refreshAction.objectUid != null) {
                   refreshWellboreObject(refreshAction);
+                } else if (
+                  (refreshAction.originJobType as JobType) ===
+                  JobType.BatchModifyObjectsOnSearch
```
eliasbruvik: Instead of checking the jobType (as I think it shouldn't be a part of RefreshAction), it would be nice if we could:

* When the RefreshAction is of type BatchModify, loop through the objects, and run invalidateObject on each individual object. This way we would not only invalidate the search, but also each object after the batch update.
* In refreshWellboreObject, we should also run refreshSearchQuery so that if an object is updated, we would always make sure that the updated values are represented in the search regardless of how the object is updated.

eliasbruvik (changes requested):
I commented a few suggestions for improvement which will reduce code duplication and avoid introducing originJobType to all the workers. As it needs to be tested, we might need to make some small adjustments in case it would lead to too many re-fetches, but I think it would be fine.

Src/WitsmlExplorer.Api/Models/RefreshAction.cs
vaclavbasniar: Removed and refresh condition changed.

Src/WitsmlExplorer.Frontend/components/RefreshHandler.tsx
vaclavbasniar: Removed and refresh condition changed.

Src/WitsmlExplorer.Api/Jobs/BatchModifyObjectsOnSearchJob.cs
vaclavbasniar: Removed and BatchModifyObjectsOnWellboreWorker updated.

eliasbruvik (changes requested):
Good work, this is exactly what I had in mind. 

There's some extra logic to avoid re-fetching some queries if all the objects are already fetched. That logic does not currently take invalidation into consideration. Can you add this function to the end of `useGetObjectSearch.tsx`, and set `cachedAllObjects` in the `useGetObjectSearch` hook to use that function?

`const cachedAllObjects = getCachedAllObjects(queryClient, server.url, filterType);`

```
const getCachedAllObjects = (queryClient: QueryClient, serverUrl: string, filterType: ObjectFilterType) => {
  const allObjectsQueryKey = getObjectSearchQueryKey(serverUrl, filterType, "", true);
  const cachedState = queryClient.getQueryState(allObjectsQueryKey)
  if (cachedState?.isInvalidated) {
    return null;
  }
  return queryClient.getQueryData<ObjectSearchResult[]>(
    allObjectsQueryKey
  );
}
```

When that is added, the invalidation will cause a re-fetch even when we already have fetched all the objects which is what we want.

eliasbruvik (approved):
- [x] Read issue and PR
- [x] Reviewed the code

Issue #2534, opened 2024-08-29
UID Mapping


vaclavbasniar:
Some companies and their systems have different names for the same wellbores. Wellbore UIDs between systems don't match. 

When opening server wellbore’s context menu sub-menu _Map UID From Server_ with list of target servers is listed. When clicked on target server, following dialog for mapping definition is shown:

![image](https://github.com/user-attachments/assets/561b6250-c43a-4e32-bc55-80adfdc60931)

The mapping will be stored in a dictionary, which could be used e.g. for _copy to server_, or _show on server_ functionality. The dictionary will be stored on a WITSML backend.

PR #2567, opened 2024-10-10
#2534 UID Mapping

vaclavbasniar:

## Fixes

This pull request fixes issue #2534

## Description

- added modal
- added storage
- updated copy and show on server functionality
- removed connected server from list of servers in context menus
- unit tests

## Type of change

* [ ] Bugfix
* [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [x] This change requires a documentation update

* [x] Frontend
* [x] API
* [ ] WITSML
* [ ] Desktop
* [ ] Other (please describe)

## Checklist:

*Communication*

*Code quality*

*Test coverage*

## Further comments


Src/WitsmlExplorer.Api/Models/UidMapping.cs
```diff
@@ -0,0 +1,81 @@
...
+        public Guid TargetServerId { get; set; }
+
+        [JsonPropertyName("targetWellId")]
+        public string TargetWellId { get; set; }
+
+        [JsonPropertyName("targetWellboreId")]
+        public string TargetWellboreId { get; set; }
+
+        [JsonPropertyName("username")]
+        public string Username { get; set; }
```
eliasbruvik, line 28: What is the intention of _Username_ for a mapping? Can't see that it's used anywhere in the frontend. Is it to keep a reference to the person who last created/updated a mapping?

Src/WitsmlExplorer.Api/Services/UidMappingService.cs
```diff
@@ -0,0 +1,189 @@
...
+
+            await _mappingRepository.UpdateDocumentAsync(new(uidMapping.SourceServerId, uidMapping.TargetServerId), mappings);
+
+            return uidMapping;
+        }
+
+        public async Task<ICollection<UidMapping>> QueryUidMapping(UidMappingDbQuery query)
+        {
+            if ((query.SourceWellId.IsNullOrEmpty() && query.SourceWellboreId.IsNullOrEmpty())
+                || (!query.SourceWellId.IsNullOrEmpty() && !query.SourceWellboreId.IsNullOrEmpty()))
```
eliasbruvik: Noticed that you don't check the server Guid's like you do in the Verify method in the handler. Is this intentional? Is there any reason why we can't validate it in the handler like we do for create/update?

Src/WitsmlExplorer.Api/Services/UidMappingService.cs
```diff
@@ -0,0 +1,189 @@
...
+using System.Collections;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Globalization;
+using System.Linq;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+using Microsoft.AspNetCore.Http;
+using Microsoft.AspNetCore.Mvc;
```
eliasbruvik: There are several unused directives in most of the .cs files. Could you remove these?

Src/WitsmlExplorer.Frontend/services/uidMappingService.tsx
```diff
@@ -0,0 +1,73 @@
...
+      undefined
+    );
+    if (response.ok) {
+      return response.json();
+    } else {
+      return [];
+    }
+  }
+
+  public static async queryRemoveUidMapping(
```
eliasbruvik: Shouldn't this be _removeUidMapping_ without _query_?

eliasbruvik: Doesn't look like this is currently used. Do you have any plans on supporting removal of mappings? What would currently happen if you incorrectly map something?

Src/WitsmlExplorer.Api/Services/UidMappingService.cs
eliasbruvik, line 1: Is the mapping bi-directional? If so, how is this currently achieved? If not, is there any reason why it shouldn't be?

Src/WitsmlExplorer.Api/Services/UidMappingService.cs
```diff
@@ -0,0 +1,189 @@
...
+            ICredentialsService credentialsService)
+        {
+            _mappingRepository = mappingRepository;
+            _witsmlServerRepository = witsmlServerRepository;
+            _credentialsService = credentialsService;
+        }
+
+        public async Task<UidMapping> CreateUidMapping(UidMapping uidMapping, HttpContext httpContext)
+        {
+            var mappings = await _mappingRepository.GetDocumentAsync(new(uidMapping.SourceServerId, uidMapping.TargetServerId));
```
eliasbruvik: Tested it with LiteDB, but got this error. It needs to work with every database we support (LiteDB, CosmosDB & MongoDB). 
image

eliasbruvik (changes requested):
Sorry for taking a while. Started on a review, but not finished yet. Anyways, added some comments on things that can be improved, and some questions to help me understand the logic.

Src/WitsmlExplorer.Api/Models/UidMapping.cs
vaclavbasniar: Yes, as a reference to a person who did the mapping. Will be shown in Uid Mapping Overview #2568

Src/WitsmlExplorer.Frontend/services/uidMappingService.tsx
vaclavbasniar: Delete is going to be implemnted in Uid Mapping Overview #2568

eliasbruvik: Makes sense, hadn't noticed that issue yet.

Src/WitsmlExplorer.Api/Services/UidMappingService.cs
vaclavbasniar: We had a discussion about this with Aker and it would be better to keep it uni-directional for now to have full control over the mapping.

Src/WitsmlExplorer.Frontend/components/ContextMenus/ContextMenuUtils.tsx
```diff
@@ -87,56 +90,96 @@ export const onClickShowObjectOnServer = async (
...
-      objectType
-    );
+    const dbQuery: UidMappingDbQuery = {
+      sourceServerId: connectedServer.id,
+      sourceWellId: objectOnWellbore.wellUid,
+      sourceWellboreId: objectOnWellbore.wellboreUid,
+      targetServerId: targetServer.id
+    };
+
+    const mappings = await UidMappingService.queryUidMapping(dbQuery);
```
eliasbruvik: Is it intentional to only use the mapping for some of the objects? I think you probably want to apply this to all the if/else cases.

Src/WitsmlExplorer.Frontend/components/ContextMenus/ContextMenuUtils.tsx
```diff
@@ -87,56 +90,96 @@ export const onClickShowObjectOnServer = async (
...
+
+  const mappings = await UidMappingService.queryUidMapping(dbQuery);
+
+  if (mappings.length > 0) {
+    targetWellId = mappings[0].targetWellId;
+    targetWellboreId = mappings[0].targetWellboreId;
+  } else {
+    targetWellId = wellbore.wellUid;
+    targetWellboreId = wellbore.uid;
+  }
```
eliasbruvik: There are several occasions where logic similar to the one you added above is used. Can you create an async function that accepts the necessary parameters and returns the mapping id's if existant, or the input id's if not?

eliasbruvik (changes requested):

Src/WitsmlExplorer.Frontend/components/ContextMenus/ContextMenuUtils.tsx
vaclavbasniar: It was intentional at first, however it could be implemented for all cases. Since this is just show on server and in the case there is no corresponding object on wellbore with matching UID on target server, information that the object doesn't exist is shown.

eliasbruvik: I mean `if (objectType === ObjectType.Log) {...` and `} else if (isExpandableGroupObject(objectType)) {...`. Currently you are not using the mapping for showOnServer for these objects: 
image

Src/WitsmlExplorer.Api/Services/UidMappingService.cs
vaclavbasniar: added validation

Src/WitsmlExplorer.Api/Services/UidMappingService.cs
vaclavbasniar: removed

Src/WitsmlExplorer.Frontend/services/uidMappingService.tsx
vaclavbasniar: changed

Src/WitsmlExplorer.Api/Services/UidMappingService.cs
vaclavbasniar: Replaced complex object with serialized JSON

Src/WitsmlExplorer.Frontend/components/ContextMenus/ContextMenuUtils.tsx
vaclavbasniar: added

Src/WitsmlExplorer.Frontend/components/ContextMenus/ContextMenuUtils.tsx
vaclavbasniar: added function for most of the cases

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyToServer.tsx
```diff
@@ -31,38 +33,63 @@ export const onClickCopyToServer = async (
...
+  let targetWellboreRef: WellboreReference;
+
+  const dbQuery: UidMappingDbQuery = {
+    sourceServerId: sourceServer.id,
+    sourceWellId: sourceWellUid,
+    sourceWellboreId: sourceWellboreUid,
+    targetServerId: targetServer.id
   };
 
+  const mappings = await UidMappingService.queryUidMapping(dbQuery);
```
eliasbruvik, line 50: Can't getTargetWellboreID also be used here? Not tested, but something like

```
const sourceWellboreName = toCopy[0].wellboreName;

  const { targetWellId, targetWellboreId } = await getTargetWellboreID({
    sourceServerId: sourceServer.id,
    sourceWellId: sourceWellUid,
    sourceWellboreId: sourceWellboreUid,
    targetServerId: targetServer.id
  });

  let wellbore: Wellbore;
  try {
    wellbore = await WellboreService.getWellbore(
      targetWellId,
      targetWellboreId,
      null,
      targetServer
    );
  } catch {
    return; // Cancel the operation if unable to authorize to the target server.
  }

  const targetWellboreRef: WellboreReference = {
    wellUid: targetWellId,
    wellboreUid: targetWellboreId,
    wellName: wellbore?.wellName ?? sourceWellName,
    wellboreName: wellbore?.name ?? sourceWellboreName
  };
```

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyToServer.tsx
eliasbruvik, line 1: If I map a wellbore to a wellbore on another server, and then delete the target wellbore, the mapping will exist but not the target wellbore. If I then try to use the copyToServer-functionality it asks me to create the target wellbore as expected. However, it fails with this message:
image

I would expect either of these (maybe this should be discussed?):
* Any mapping that exists for a wellbore that is deleted should also be removed. If this is chosen, we need to take into account what should happen if the wellbore is deleted from the query view or another application (when we don't know that it has been deleted).
* CopyToServer will create the wellbore with the mapped uid, and then copy the object over (without any errors).

eliasbruvik (changes requested):

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyToServer.tsx
vaclavbasniar: Yes, that's a good point, that hasn't been discussed. I think, the first solution is correct. We should probably discuss this in weekly meeting.

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyToServer.tsx
vaclavbasniar: Yes, I'll check that again. However, I guess this would probably need a little bit more refactoring.

Src/WitsmlExplorer.Frontend/components/Modals/WellboreUidMappingModal.tsx
eliasbruvik, line 1: Tested how this would work on a production server that has a lot of wells/wellbores. It completely freezes the application for about 30 sec whenever I make a small change to the search. 

It's a bit laggy (~1 sec) when changing a search with few matches, and very laggy when there are a lot of results, so this might mostly be a rendering issue. Can you try to add a small debounce to the input to avoid unnecessary full re-renders, and to virtualize the treeview?

eliasbruvik (changes requested):

Src/WitsmlExplorer.Frontend/components/Modals/WellboreUidMappingModal.tsx
vaclavbasniar: Yes, I believe, that it's the rendering issue. I'll add the virtualization and debounce.

vaclavbasniar: Is there a possibility we could get a test server with such a setup - lot of well, wellbores, data in general? I think it could be beneficial also for future development cases.

eliasbruvik: I agree that it would be beneficial, but it hasn't been easy to get access to "public" test servers in the past. We had some discussions in the past on getting more realistic datasets over to the public servers, but that has been downprioritized for a while. If we want more data, I think we would have to generate a lot of wells/wellbores on one of the public servers ourselves.

vaclavbasniar: Added debounce, virtualization and some small performance improvements.

eliasbruvik (dismissed):
*Reviewer checklist*
- [x] Read issue and PR
- [x] Reviewed the code

eliasbruvik (approved):

PR #2667, opened 2025-03-05
Added optional CancellationToken argument to GetFromStore methods #2666

### LINKED ISSUE #2666: Add support/use of CancellationToken as argument to GetFromStore methods
Clients of the Witsml library should be able to pass a CancellationToken to GetFromStore requests to cancel tasks that are running. This can be either due to application shutdowns, test timeouts, or other usages.

I've looked at the `dotnet-svcutil` which is the utility that is used for generating the soap client, and it does not seem to support generating methods with this directly. It also does not look like SOAP supports cancellation.

Solving this would thus be restricted to cancelling the tasks within the client itself so that the client is released although the requests at the server level will continue.

It only make sense to add this to the GetFromStore methods since the server requests will continue, so AddToStore, UpdateInStore and DeleteFromStore should remain as is

lindsve:
## Fixes

This pull request fixes #2666

## Description

It adds an optional `CancellationToken` argument to the GetFromStore methods in `WitsmlClient` and implements cancelling of the task in `WitsmlMetrics.MeasureQuery()`

## Type of change

* [ ] Bugfix

* [ ] Frontend
* [ ] API
* [x] WITSML
* [ ] Desktop
* [ ] Other (please describe)

## Checklist:

*Communication*

*Code quality*

*Test coverage*

### DIFF
--- a/Src/Witsml/Metrics/WitsmlMetrics.cs
+++ b/Src/Witsml/Metrics/WitsmlMetrics.cs
@@ -52,14 +52,13 @@ internal sealed class WitsmlMetrics
             { "objectType", witsmlType },
         };
 
-        var ct = cancellationToken ?? CancellationToken.None;
         Stopwatch timer = null;
         TResponseType response;
         try
         {
             _activeRequests.Add(1, tagList);
             timer = Stopwatch.StartNew();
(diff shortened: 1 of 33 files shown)


robertbasti (approved):
seems ok

PR #2686, opened 2025-03-29
Copy whole wellbore

### LINKED ISSUE #2687: Copy whole wellbore - basic implementation
Implement happy day scenario  for copying all wellbore objects when wellbore with the same name /uid does not exists on the target.

robertbasti:

## Fixes

This pull request fixes #2687

## Description

Basic implementation of copying a wellbore with all objects. If wellbore already exists on the target, will not be overwritten at that stage.

## Type of change

* [ ] Bugfix

* [x] Frontend
* [x] API
* [x] WITSML
* [ ] Desktop
* [ ] Other (please describe)

## Checklist:

*Communication*

*Code quality*

*Test coverage*

## Further comments


Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -0,0 +1,253 @@
...
+
+            BaseReport report = CreateCopyWithParentReport(reportItems);
+            job.JobInfo.Report = report;
+
+            RefreshAction refreshAction = new RefreshWell(targetClient.GetServerHostname(), job.Target.WellUid, RefreshType.Update);
+            wellboreResult.refresh = refreshAction;
+            return wellboreResult;
+        }
+
+        private CommonCopyReport CreateCopyWithParentReport(List<CommonCopyReportItem> reportItems)
```
eliasbruvik: Rename this to CreateCopyWellboreWithObjectsReport?

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -0,0 +1,253 @@
...
+                Source = new WellboreReference()
+                {
+                    WellUid = job.Source.WellUid,
+                    WellboreUid = job.Source.WellboreUid,
+                    WellName = job.Source.WellName,
+                    WellboreName = job.Source.WellboreName
+                }
+            };
+            (WorkerResult result, RefreshAction refresh) wellboreResult =
+                await _copyWellboreWorker.Execute(copyWellboreJob,
```
eliasbruvik, line 71: If this wellborecopyjob fails, won't it still continue executing the consecutive copy jobs? I think you need to check wellboreResult.result.IsSuccess to know if you should continue or not.

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -0,0 +1,253 @@
...
+                cancellationToken);
+
+            await CopyLogs(job, reportItems, sourceClient, cancellationToken);
+
+            BaseReport report = CreateCopyWithParentReport(reportItems);
+            job.JobInfo.Report = report;
+
+            RefreshAction refreshAction = new RefreshWell(targetClient.GetServerHostname(), job.Target.WellUid, RefreshType.Update);
+            wellboreResult.refresh = refreshAction;
+            return wellboreResult;
```
eliasbruvik: Can you create a new WorkerResult instead of using the result from the other job? Then use a message with something like "Successfully copied wellbore with all supported child objects: {job.Source.WellboreUid} -> {job.Target.WellboreUid}". Or alternatively if some jobs failed: "Successfully copied wellbore with x of y supported child objects: {job.Source.WellboreUid} -> {job.Target.WellboreUid}".
Also, if some of the child copy jobs fails, should the entire job end up as a success or not?

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -0,0 +1,253 @@
...
+                reportItems, cancellationToken);
+            await CopyWellboreObjectsByType(job, EntityType.Trajectory,
+                sourceClient, reportItems, cancellationToken);
+            await CopyWellboreObjectsByType(job, EntityType.Tubular,
+                sourceClient, reportItems, cancellationToken);
+            await CopyWellboreObjectsByType(job,
+                EntityType.WbGeometry, sourceClient, reportItems,
+                cancellationToken);
+
+            await CopyLogs(job, reportItems, sourceClient, cancellationToken);
```
eliasbruvik: Is it necessary to use a different method for logs? Doesn't the CopyObjectsWorker make sure that the logs are handled correctly?
image

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
eliasbruvik: Actually, it seems like if the wellbore already exists, the CopyWellboreWorker considers that a success, which would mean that the job would actually be successful and we would copy all the objects over to the existing wellbore. I don't see why that should be a success, so maybe you could look into the other jobs that uses it to see if you can change it to failed?

image

eliasbruvik (changes requested):

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: done

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: done

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: done

robertbasti: at least at that phase - partial fail could be also success to see the report. Needs a discussion with application users.

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: fixed

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -64,6 +65,19 @@ public CopyWellboreWithObjectsWorker(ILogger<CopyWellboreWithObjectsJob> logger,
             (WorkerResult result, RefreshAction refresh) wellboreResult =
                 await _copyWellboreWorker.Execute(copyWellboreJob,
                     cancellationToken);
+            string wellboreAlredyExists = "Target wellbore already exists";
+            if (wellboreResult.result.Message == wellboreAlredyExists)
```
eliasbruvik: It's a bit dangerous to check if it already exists this way. In case that string is changed from the worker at a later point, they won't know that they also need to check it here. And if that happens, we can cause a lot of problems. 

Instead, I propose to just run the query to manually check if the wellbore exists directly in this worker. For now, if it already exists, just fail the job. Later we can handle that case with some extra logic to merge objects or only copy objects that's not already on the target.

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -91,17 +105,33 @@ await CopyWellboreObjectsByType(job,
...
+            BaseReport report = CreateCopyWellboreWithObjectsReport(reportItems);
             job.JobInfo.Report = report;
 
-            RefreshAction refreshAction = new RefreshWell(targetClient.GetServerHostname(), job.Target.WellUid, RefreshType.Update);
-            wellboreResult.refresh = refreshAction;
-            return wellboreResult;
+            var fails = reportItems.Count(x => x.Status == "Fail");
+            if (fails > 0)
+            {
+                WorkerResult workerResult = new(targetClient.GetServerHostname(), true, $"Successfully copied wellbore with supported child objects, copying of some objects failed: {job.Source.WellboreUid} -> {job.Source.WellboreUid}", sourceServerUrl: sourceClient.GetServerHostname());
```
eliasbruvik: Could this be a better message? "Partially copied wellbore with some child objects. Failed to copy {fails} out of {reportItems.Count()} objects."

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -91,17 +105,33 @@ await CopyWellboreObjectsByType(job,
...
+                RefreshAction refreshAction = new RefreshWell(targetClient.GetServerHostname(), job.Target.WellUid, RefreshType.Update);
+                return (workerResult, refreshAction);
+            }
+            else
+            {
+                WorkerResult workerResult = new(targetClient.GetServerHostname(), true, $"Successfully copied wellbore with all supported child objects: {job.Source.WellboreUid} -> {job.Source.WellboreUid}", sourceServerUrl: sourceClient.GetServerHostname());
```
eliasbruvik: `{job.Source.WellboreUid} -> {job.Source.WellboreUid}`
Does this part actually provide any information? The uids are hard to read for the user anyways.

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -148,94 +172,88 @@ private async Task CopyLogs(CopyWellboreWithObjectsJob job, List<CommonCopyRepor
...
-                        {
-                            WellUid = job.Source.WellUid,
-                            WellName = job.Source.WellName,
-                            WellboreUid = job.Source.WellboreUid,
-                            WellboreName = job.Source.WellboreUid,
-                            ObjectType = entityType,
-                            ObjectUids = result.Objects.Select(x => x.Uid).ToArray()
-                        },
-                        Target = new WellboreReference()
+                        foreach (var timeLog in result.Objects)
```
eliasbruvik: As agreed in a call, we should loop through all objects one at a time to make sure that we continue on the next one if the previous fails.

eliasbruvik (changes requested):

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyUtils.tsx
```diff
@@ -127,6 +138,43 @@ export const pasteComponents = async (
...
 
+export const pasteWellbore = async (
+  servers: Server[],
+  wellBore: WellboreReference,
+  dispatchOperation: DispatchOperation,
+  well: Well,
+  connectedServer: Server
+) => {
+  dispatchOperation({ type: OperationType.HideContextMenu });
+  const existingWellbores = await WellboreService.getWellbores(
```
eliasbruvik: I think this validation of existing wellbores should be handled in the backend. The backend should just return a failed job with a proper message, which will then be shown to the user. So just remove this check and the alertDispatcher dispatch from this function.

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyUtils.tsx
```diff
@@ -127,6 +138,43 @@ export const pasteComponents = async (
   onClickPaste(servers, sourceReferences?.serverUrl, orderCopyJob);
 };
 
+export const pasteWellbore = async (
+  servers: Server[],
+  wellBore: WellboreReference,
```
eliasbruvik: Can you rename this to sourceWellbore?

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyUtils.tsx
```diff
@@ -127,6 +138,43 @@ export const pasteComponents = async (
...
+  dispatchOperation: DispatchOperation,
+  well: Well,
```
eliasbruvik: Can you rename this to targetWell?

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
```diff
@@ -34,6 +35,39 @@ export const useClipboardReferences = (
...
+  const [wellboreReference, setWellboreReference] =
+    useState<WellboreReference>(null);
+
+  const tryToParseClipboardContent = async () => {
+    try {
+      const clipboardText = await navigator.clipboard.readText();
+      const wellboreReference = parseWellboreStringToReference(clipboardText);
+      setWellboreReference(wellboreReference);
+    } catch (e) {
+      throw new Error("Invalid input given.", e);
```
eliasbruvik: Why do you throw an error here?

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
```diff
@@ -34,6 +35,39 @@ export const useClipboardReferences = (
   return objectReferences;
 };
 
+export const useWellBoreReference = (
```
eliasbruvik: Lowercase b for Wellbore.

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
```diff
@@ -34,6 +35,39 @@ export const useClipboardReferences = (
...
+  pollInterval = 0
+): WellboreReference | null => {
+  const [wellboreReference, setWellboreReference] =
+    useState<WellboreReference>(null);
+
+  const tryToParseClipboardContent = async () => {
+    try {
+      const clipboardText = await navigator.clipboard.readText();
+      const wellboreReference = parseWellboreStringToReference(clipboardText);
+      setWellboreReference(wellboreReference);
```
eliasbruvik, line 47: I see it's done the same way for objects, but what happens if you first copy a real wellbore reference, and then copy something else? It does not look like we ever set the reference back to null?

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: fixed

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
```diff
@@ -53,6 +87,18 @@ export function parseStringToReferences(input: string): ObjectReferences {
...
 }
 
+export function parseWellboreStringToReference(
+  input: string
+): WellboreReference {
+  let jsonObject: WellboreReference;
+  try {
+    jsonObject = JSON.parse(input);
+  } catch (e) {
+    throw new Error("Invalid input given.", e);
```
eliasbruvik, line 95: I don't think JSON.parse actually checks that the types corrsepond to the jsonObject you are placing it. It only fails if it's not valid JSON. I think you need to do manual checks on object and ensure that the correct parameters (and only those) exists. 

As seen here, it did not throw an error when I copied an object (not a wellbore):

image

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: fixed

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: fixed

Src/WitsmlExplorer.Frontend/models/jobs/wellboreReference.ts
```diff
@@ -1,10 +1,12 @@
+import { Server } from "models/server";
 import Wellbore from "models/wellbore";
 
 export default interface WellboreReference {
   wellUid: string;
   wellboreUid: string;
   wellName: string;
   wellboreName: string;
+  server?: Server;
```
eliasbruvik: Can you use a serverUrl string instead of the entire server object?

eliasbruvik (changes requested):

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: fixed

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyUtils.tsx
robertbasti: fixed

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyUtils.tsx
robertbasti: done

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyUtils.tsx
robertbasti: done

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
robertbasti: fixed

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
robertbasti: fixed

Src/WitsmlExplorer.Frontend/models/jobs/wellboreReference.ts
robertbasti: fixed

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
eliasbruvik: You should check if it exists before you try to copy the wellbore, not after.

robertbasti: fixed

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
robertbasti: fixed with a new properties check

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -0,0 +1,219 @@
...
+            return new CommonCopyReport
+            {
+                Title = $"Copy wellbore with objects report",
+                Summary = "Copy wellbore  parent report",
+                ReportItems = reportItems,
+                JobDetails = "report details"
+            };
+        }
+
+        private async Task CopyOneObject(WitsmlObjectOnWellbore objectOnWellbore, CopyWellboreWithObjectsJob job, List<CommonCopyReportItem> reportItems, CancellationToken? cancellationToken)
```
eliasbruvik: Instead of adding report items to the list in this function, can you make it return the CommonCopyReportItem? It's easier to understand functions if they don't mutate/change objects directly. Then we also get information from the return type of what to expect back.

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyUtils.tsx
```diff
@@ -127,6 +138,26 @@ export const pasteComponents = async (
   onClickPaste(servers, sourceReferences?.serverUrl, orderCopyJob);
 };
 
+export const pasteWellbore = async (
+  servers: Server[],
+  sourceWellBore: WellboreReference,
```
eliasbruvik: Lowercase b here also.

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
eliasbruvik: You removed the wrong error. Remove the one at line 50 instead, the one you removed can be kept.

eliasbruvik (changes requested):

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
eliasbruvik: Do the same for the CopyWellboreObjectsByType function. It should return the list of report items, and then add all those to your original list in the execute function.

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
robertbasti: alredy removed (replaced by console.error)

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: fixed

Src/WitsmlExplorer.Frontend/components/ContextMenus/CopyUtils.tsx
robertbasti: fixed

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
```diff
@@ -93,10 +93,34 @@ export function parseWellboreStringToReference(
...
+    "serverUrl"
+  ];
+  const hasRequiredProperties = requiredProps.every((prop) =>
+    Object.prototype.hasOwnProperty.call(jsonObject, prop)
+  );
+  if (!hasRequiredProperties) {
+    throw new Error("Missing required wellbore fields.");
+  }
+  Object.keys(jsonObject).forEach((key) => {
+    if (Object.prototype.hasOwnProperty.call(jsonObject, key)) {
```
eliasbruvik: You are looping over the keys, and then checking if the key exists, so this will always throw your error. Is this what you are trying to do?

```
Object.keys(jsonObject).forEach((key) => {
    if (!requiredProps.includes(key)) {
      throw new Error(`Unexpected property found: ${key}`);
    }
  });
  ```

File: Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs [hunk 20]
```diff
@@ -0,0 +1,227 @@
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.AspNetCore.Mvc;
```
eliasbruvik: Not used

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -0,0 +1,227 @@
...
+{
+    public interface ICopyWellboreWithObjectsWorker
+    {
+        Task<(WorkerResult, RefreshAction)> Execute(CopyWellboreWithObjectsJob job, CancellationToken? cancellationToken = null);
+    }
+
+    public class CopyWellboreWithObjectsWorker : BaseWorker<CopyWellboreWithObjectsJob>, IWorker, ICopyWellboreWithObjectsWorker
+    {
+        private readonly ICopyWellboreWorker _copyWellboreWorker;
+        private readonly ICopyLogWorker _copyLogWorker;
```
eliasbruvik: Not used anymore

eliasbruvik (changes requested):

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
robertbasti: fixed by last commit

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: fixed

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
robertbasti: fixed

Src/WitsmlExplorer.Frontend/components/ContextMenus/UseClipboardReferences.ts
eliasbruvik: Nvm, these hooks are only used within the context menus, so if you leave it, copy something else, and then return, it will start from scratch at null.

Src/WitsmlExplorer.Api/Workers/Copy/CopyWellboreWithObjectsWorker.cs
```diff
@@ -0,0 +1,224 @@
...
+                        WellboreUid = job.Source.WellboreUid,
+                        WellName = job.Target.WellName,
+                        WellUid = job.Target.WellUid,
+                    }
+                };
+                (WorkerResult result, RefreshAction refresh) copyObjectResult = await _copyObjectsWorker.Execute(copyLogJob, cancellationToken);
+                var reportItem = new CommonCopyReportItem()
+                {
+                    Phase = "Copy " + objectOnWellbore.Name,
+                    Message = copyObjectResult.result.Message,
```
eliasbruvik, line 186: Can you create a new CopyWellboreWithObjectsReportItem class based on the common one, but that also contains a "ObjectType" string that will be set to the object type? In the API you could take the values from EntityType. This way it's easy to filter on specific object types in the report modal.

eliasbruvik (changes requested):

eliasbruvik (approved):
*Reviewer checklist*
- [x] Read issue and PR
- [x] Reviewed the code

PR #2834, opened 2025-10-09
Add support for DWO in the frontend

eliasbruvik:
Initial support for DWO in the frontend. It's bare-bones, just adding DWO to ObjectType and ensuring that all mappings from ObjectType are set. Further listViews and contextMenus should be added later.

### DIFF
--- a/Src/WitsmlExplorer.Frontend/components/ContentViews/DataWorkOrderListView.tsx
+++ b/Src/WitsmlExplorer.Frontend/components/ContentViews/DataWorkOrderListView.tsx
@@ -1,0 +1,115 @@
+import {
+  ContentTable,
+  ContentTableColumn,
+  ContentTableRow,
+  ContentType
+} from "components/ContentViews/table";
+import formatDateString from "components/DateFormatter";
+import { useConnectedServer } from "contexts/connectedServerContext";
+import { useGetObjects } from "hooks/query/useGetObjects";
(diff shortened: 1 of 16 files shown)


Src/WitsmlExplorer.Frontend/components/ContentViews/DataWorkOrderListView.tsx
```diff
@@ -0,0 +1,115 @@
...
+  );
+
+  useExpandSidebarNodes(wellUid, wellboreUid, ObjectType.DataWorkOrder);
+
+  const getTableData = () => {
+    return dataWorkOrders.map((dataWorkOrder) => {
+      return {
+        ...dataWorkOrder,
+        ...dataWorkOrder.commonData,
+        id: dataWorkOrder.uid,
```
robertbasti, line 40: should not be it uid?

robertbasti (changes requested):

Src/WitsmlExplorer.Frontend/components/ContentViews/DataWorkOrderListView.tsx
eliasbruvik: No, the uid is already spread with `...dataWorkOrder`. The 'id' must be set for all data rows sent to ContentTable because it's used by React and Tanstack Table to have unique ids for each row for (re)-rendering purposes.

robertbasti (approved):

PR #2835, opened 2025-10-09
Add DWO where all entity types are expected

eliasbruvik:
Added DWO to tests and other locations where we expect to use all entity types


robertbasti (approved):
seems ok

PR #2837, opened 2025-10-13
DWO Split queries and add some basic views

eliasbruvik:


Src/WitsmlExplorer.Api/Query/DataWorkOrderQueries.cs
```diff
@@ -40,97 +42,13 @@ public static WitsmlDataWorkOrders GetWitsmlDataWorkOrder(string wellUid, string
...
                     }
                 ],
-                DataSourceConfigurationSets = new List<WitsmlDataSourceConfigurationSet>()
-                {
-                    new WitsmlDataSourceConfigurationSet()
+                DataSourceConfigurationSets =
+                [
+                    new()
                     {
-                        Uid = "",
```
robertbasti, line 47: we don't need the whole structure any more ?

robertbasti (changes requested):

Src/WitsmlExplorer.Api/Query/DataWorkOrderQueries.cs
eliasbruvik: No, if we need those "deeper" elements, we'll just simplify it by using the QueryById with ReturnElements All. Then we'll get every nested element without specifying it. 

We could theoretically always do it that way, but for the "short" version we don't need anything within a set, so then we save a bit by only requesting what we need.

robertbasti (approved):

